PR #11, opened 2023-05-09
Feat/milestone1

aymericdelab:
**1st milestone containing:**
1. simplified ERC721 for realms with mint, transfer, approve and the ability to store realms data and position
2. settling: Settle, Unsettle
3. labor: Build Labor, Harvest Labor (for food and non food resources)

All tested as well.


src/components/position.cairo
```diff
@@ -0,0 +1,25 @@
...
+
+impl PositionImpl of PositionTrait {
+    // fn is_zero(self: Position) -> bool {
+    //     match self.x - self.y {
+    //         0 => false,
+    //         _ => true,
+    //     }
+    // }
+
+    fn is_equal(self: Position, b: Position) -> bool {
```
milancermak, line 22: A better approach would be to impl `PartialEq` so we can then do `==` and `!=` directly.

If we need `is_zero`, you can also impl `Zeroable`.

src/components/realm.cairo
```diff
@@ -0,0 +1,59 @@
...
+    fn is_owner(self: Realm, address: ContractAddress) -> bool {
+        if (self.owner == address) {
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    fn has_resource(self: Realm, resource_id: u8) -> bool {
+        let mut resource_ids: Array<u8> = unpack_resource_ids(
```
milancermak, line 44: Wouldn't it be better if `unpack_resource_ids` is a method on self? Does it make sense outside of a Realm component?

Now I'm even starting to think it would be cool to have some kind of lazily loaded component members - I can imagine `has_resource` can be called multiple times in a single TX, which would call `unpack_resource_ids` N times as well, but it's really only necessary to do it once. Not sure if it's possible yet in Cairo/Dojo, but an intriguing idea nonetheless. 

Also, in any case, the `unpack_resource_ids` should IMO return a `Span<u8>`, not an array, as we don't need to modify it after unpacking, do we?

src/components/realm.cairo
```diff
@@ -0,0 +1,59 @@
...
+            self.resource_ids_packed, self.resource_ids_count
+        );
+        let mut index = 0_usize;
+        let has_resource = loop {
```
milancermak, line 48: So there's an [more efficient way of looping through an array](https://twitter.com/PapiniShahar/status/1650814998480838656) using `pop_front` - there's no need for a counter variable and you save a RangeCheck builtin on every iteration. Coincidentally I have a post about it coming out tomorrow on Cairopractice 😄

Would be cool if you can refactor this bit to use it.

src/components/tick.cairo
```diff
@@ -0,0 +1,18 @@
+use eternum::components::config::WorldConfig;
+use eternum::utils::math::get_past_time;
```
milancermak, line 2: This is missing.

src/components/tick.cairo
```diff
@@ -0,0 +1,18 @@
...
+use eternum::constants::WORLD_CONFIG_ID;
+
+#[derive(Component)]
+struct Tick {
+    last_update: u128
+}
+trait TickTrait {
+    fn needs_update(self: Tick, world_config: WorldConfig, block_timestamp: u128) -> bool;
```
milancermak, line 10: `block_timestamp` should be `u64`

src/components/tick.cairo
```diff
@@ -0,0 +1,18 @@
+use eternum::components::config::WorldConfig;
+use eternum::utils::math::get_past_time;
+use eternum::constants::WORLD_CONFIG_ID;
```
milancermak, line 3: Not used.

src/constants.cairo
```diff
@@ -0,0 +1,57 @@
...
+const WORLD_CONFIG_ID: felt252 = 999999999999999999;
+const BUILDING_CONFIG_ID: felt252 = 999999999999999998;
+const LABOR_CONFIG_ID: felt252 = 999999999999999997;
+
+// 8 bits
+const RESOURCE_IDS_PACKED_SIZE: usize = 8_usize;
+const REALMS_DATA_PACKED_SIZE: usize = 8_usize;
+
+// prime 
+const PRIME: felt252 = 3618502788666131213697322783095070105623107215331596699973092056135872020480;
```
milancermak, line 11: To be fair, this is max(felt252), not the Cairo prime. It doesn't seem to be used anywhere either.

src/constants.cairo
```diff
@@ -0,0 +1,57 @@
...
+    const CURIOUS_SPORE: u8 = 25;
+    const UNREFINED_ORE: u8 = 26;
+    const SUNKEN_SHEKEL: u8 = 27;
+    const DEMONHIDE: u8 = 28;
+    const WHEAT: u8 = 254;
+    const FISH: u8 = 255;
+}
+
+
+enum BuildingIds {
```
milancermak, line 47: Why are ResourceIds a mod with const but BuildingIds an enum?

src/systems/buildings.cairo
```diff
@@ -0,0 +1,177 @@
+// TODO for a later milestone
```
milancermak, line 1: Note: skipping this file from review for now then.

src/systems/config/labor_config.cairo
```diff
@@ -0,0 +1,65 @@
+#[system]
+mod CreateLaborCR {
+    use traits::Into;
+    use traits::TryInto;
+    use array::ArrayTrait;
+    use box::BoxTrait;
+    use eternum::utils::unpack::unpack_resource_ids;
+    use eternum::components::config::LaborCR;
+    use eternum::components::config::LaborCV;
```
milancermak, line 9: BTW we can now group imports, so `use eternum::components::config::{LaborCR, LaborCV}`.

src/components/config.cairo
```diff
@@ -0,0 +1,63 @@
...
+#[derive(Component)]
+struct LaborConf {
+    base_labor_units: u128, // 86400 / 12    
+    vault_percentage: u128, // 250 
+    base_resources_per_cycle: u128, // (252 / 12) * 10 ** 18;
+}
+
+// labor cost resources
+#[derive(Component)]
+struct LaborCR {
```
milancermak, line 49: Would prefer a more verbose name. When viewing in code, meanings of `CR` and `CV` are somewhat difficult to remember and also too similar.

src/systems/config/labor_config.cairo
```diff
@@ -0,0 +1,65 @@
...
+
+    use debug::PrintTrait;
```
milancermak, line 11: Forgotten.

src/systems/config/labor_config.cairo
```diff
@@ -0,0 +1,65 @@
+#[system]
+mod CreateLaborCR {
+    use traits::Into;
+    use traits::TryInto;
```
milancermak, line 4: Seems most of the imports are not necessary.

src/systems/labor/build_labor.cairo
```diff
@@ -0,0 +1,492 @@
...
+    use eternum::constants::LABOR_CONFIG_ID;
+    use eternum::constants::ResourceIds;
+    use eternum::utils::unpack::unpack_resource_ids;
+
+    use debug::PrintTrait;
+
+    #[external]
+    fn execute(realm_id: felt252, resource_id: u8, labor_units: u128, multiplier: u128) {
+        // assert owner of realm
+        let player_id: ContractAddress = starknet::get_caller_address();
```
milancermak, line 29: This var is immediately shadowed without being used.

FWIW I think `get_caller_address` is what we want, not `tx_info.acconut_contract_address`, but I could be wrong.

milancermak (changes requested):
Big PR, a lot of progress! Thanks for that 👑 

I didn't go through all of it yet, too tired RN to focus, so I'll continue once I'm fresh again. Please consider most of the comments as points of discussion. I don't claim to be right, I'm also interested in your point of view and in general building the best possible solution.

I've noticed the ID alias, but couple of _id variables don't use that, I didn't point all of them, but I think it's a good idea, to have a consistent type for IDs! I'm for applying it throughout the codebase.

src/alias.cairo
```diff
@@ -0,0 +1 @@
+type ID = felt252;
```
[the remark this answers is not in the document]
aymericdelab: commands::uuid() was originally returning a felt252, but now it's returning a usize so i'm thinking that having all IDs being usize as well might be better, wdyt?

src/systems/config/labor_config.cairo
```diff
@@ -0,0 +1,65 @@
...
+        );
+    }
+}
+
+#[system]
+mod CreateLaborCV {
+    use traits::Into;
+    use traits::TryInto;
+    use array::ArrayTrait;
+    use box::BoxTrait;
```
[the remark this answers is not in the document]
aymericdelab: ok will clean up all imports in the repo !

src/components/buildings.cairo
```diff
@@ -0,0 +1,36 @@
+//
+// ---------- Buildings
+// NB: Have left this open and not specifcially tied to only Realms. Barbarians could share the same buildings.
+use array::ArrayTrait;
+
+#[derive(Component)]
+struct Buildings {
```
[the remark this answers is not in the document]
aymericdelab: there's a quantity field in the Buildings component so it can hold more than one building

src/components/config.cairo
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#[derive(Component)]
+struct BuildingConfig {
+    base_sqm: u128,
+    workhut_cost: u128,
+}
+
+#[derive(Component)]
+struct LaborConf {
```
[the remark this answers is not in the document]
aymericdelab: that's something i should have mentioned but the components cannot have more than a certain number of characters du to dojo, need to create an issue in dojo

src/components/position.cairo
aymericdelab: Interesting ! will look into that !

src/components/realm.cairo
```diff
@@ -0,0 +1,59 @@
...
+    harbors: u8,
+    rivers: u8,
+    regions: u8,
+    wonder: u8,
+    order: u8
+}
+
+
+trait RealmTrait {
+    fn is_owner(self: Realm, address: ContractAddress) -> bool;
```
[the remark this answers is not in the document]
aymericdelab: agreed !

src/components/realm.cairo
aymericdelab: thought about this as well but i'm using this in other parts as well for ```LaborCR``` and ```BuildingTypeConfig``` components.
Love the idea as well but not sure if possible in Cairo. In that case i would maybe split the function into two :
1. ```get_resources``` retrieves array
2. ```has_resources``` takes array as input

Agreed on the span !

src/components/realm.cairo
aymericdelab: amazing, will try it out

src/constants.cairo
aymericdelab: my bad, forgot to remove. is prime = 2**251 + (17*2**192) + 1 still true ?

src/constants.cairo
aymericdelab: wanted to use enums for resource ids at first but since im packing them i need a number for them

src/systems/labor/build_labor.cairo
aymericdelab: thought i deleted this, sorry
basically get_caller_address will return the world address in dojo. so we need account_contract_address to get the contract calling the world. here's the related discussion : https://discord.com/channels/1062934010722005042/1062934060898459678/1103662308405805188

src/utils/unpack.cairo
```diff
@@ -0,0 +1,50 @@
...
+use traits::Into;
+use traits::TryInto;
+use option::OptionTrait;
+use eternum::constants::RESOURCE_IDS_PACKED_SIZE;
+use eternum::constants::PRIME;
+use traits::BitAnd;
+use eternum::utils::math::pow;
+use debug::PrintTrait;
+
+fn unpack_resource_ids(resource_ids_packed: u128, resource_ids_count: u8) -> Array<u8> {
```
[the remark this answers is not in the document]
aymericdelab: that was my original plan but i ran into some withdraw_gas errors when running tests and couldn't get rid of it. since withdraw_gas shouldn't be added anymore to the code and adding it does not remove the error, i'm thinking these are bugs that might get resolved in future cairo versions. will can try again in future milestones if theres more unpacking to do

src/components/army.cairo
```diff
@@ -0,0 +1,58 @@
...
+//
+// ----------- Army
+// Armies can be attacked to any Entity if it has a Spawn function. Realms can spawn them and have ownership over them, however they can exist
+// outside of a Realm. They might spawn randomly on the Map.
+
+#[derive(Component)]
+struct Army {
+    owner_entity_id: u32, // Armies are Entities. We assign an Army to a Realm for example, to give ownership to the Realm.
+    light_cavalry_qty: u32,
+    light_cavalry_health: u32,
```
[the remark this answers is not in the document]
aymericdelab: good question, i think that if we need to create a struct, maybe it means that the abstraction is too high it should be its own component?

src/components/config.cairo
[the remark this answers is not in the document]
aymericdelab: tested it again and it seems to be working, must have been a change in dojo then, am changing:

- LaborConf -> LaborConfig
- LaborCR -> LaborCostResources (packed list of resources needed to build labor on one resource)
- LaborCV -> LaborCostAmount (amount for each resourceId needed to build labor on one resource)

src/components/realm.cairo
aymericdelab: so i tested it but it does not work, the reason is that when dojo parses the code, it automatically removes the semi-colon at the end of the match block, which then gives me a ```Tail expression not allow in a `loop` block.``` error. I'm opening an issue in the dojo repo.

src/alias.cairo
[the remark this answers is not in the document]
milancermak: FWIW, I think felt252 is fine.

I actually don't like Dojo's use of `usize` there for 2 reasons: 1) UUID is defined in a RFC as 128 bits so if we're using the UUID name (even though not the implementation details), it should be `u128` and 2) `usize` is platform dependent, kind of, and I'm not really sure how this will play out with Starknet or other L3s the contracts might be deployed on in the future. It could be aliased to u16 or u256, who knows? It's the uncertainty of it that bothers me.

So I think felt252 or u128 are good options for ID.

src/components/army.cairo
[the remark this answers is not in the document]
milancermak: How would that look? Can you expand on it a little please?

src/components/config.cairo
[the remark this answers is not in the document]
milancermak: Heh, I think it was me who fixed that Dojo issue 😃

src/components/realm.cairo
milancermak: Haha, such a weird issue. Great job on investigating!

src/components/tick.cairo
milancermak: Out of scope, resolving.

src/components/tick.cairo
milancermak: Out of scope, resolving.

src/constants.cairo
milancermak: Yes, still is.

src/constants.cairo
milancermak: Can we then use `mod BuildingIds` with consts as well? I'd rather have this consistent.

milancermak: Oh, it's also out of scope for this PR right?

src/systems/labor/build_labor.cairo
milancermak: Thanks for pointing me to that conversation 🙏 Prompted by this review I asked a similar question in the cairo-1 discord, wasn't sure about the diff between those two. You're absolutely right, `account_contract_address` is the one we need 👍

src/utils/unpack.cairo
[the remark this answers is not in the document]
milancermak: Not sure if related, but I ran into a similar issue today on a different project. The "fix" was to split the test function into a couple of smaller ones, then the error went away 🤷 

FWIW, the error was:

```sh
Error: Failed setting up runner.

Caused by:
    Failed calculating gas usage, it is likely a call for `gas::withdraw_gas` is missing.
```

src/systems/settling.cairo
```diff
@@ -0,0 +1,289 @@
...
+    use eternum::components::realm::Realm;
+    use eternum::components::resources::Resource;
+    use eternum::components::age::Age;
+    use eternum::components::config::WorldConfig;
+
+    fn execute(realm_id: felt252) { // get the ERC721 contract
+        // get the owner
+        let config = commands::<WorldConfig>::entity(WORLD_CONFIG_ID.into());
+        let token = config.realm_l2_contract;
+        let owner = commands::<Owner>::entity((token, realm_id).into());
```
milancermak, line 20: Why are we getting Owner here? Wouldn't the address be 0 on settling? Besides that, why do we even need it if the Realm component has a `owner` member? Or maybe it's the the other way around, should the Realm component have `owner` at all? Wouldn't it be better to store it in a separate component, the `Owner` here?

src/utils/unpack.cairo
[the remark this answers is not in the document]
ponderingdemocritus: yeah we can tweak in next milstone, lets not block here on some upstream thing

src/systems/settling.cairo
ponderingdemocritus: agreed - If we are storing Owner in a separate component, then it's not needed on the Realm component

src/components/army.cairo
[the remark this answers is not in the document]
aymericdelab: for example you'd have the an Army component like this:
```
struct Army {
        army_type: u8,
        qty: u32,
        health: u8,
}
```
and an army would be spawned by creating an entity with keys (realm_id, army_id, army_type).

To go even further you could:
- add `Speed` component to the army
- use `Health` component instead of health field in `Army`
- add `Attack` component

src/alias.cairo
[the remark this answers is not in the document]
aymericdelab: agreed, let's use u128 then

src/constants.cairo
aymericdelab: out of scope but will add a comment

src/systems/settling.cairo
aymericdelab: agreed, let's remove the owner field and only keep `Owner` component !

src/alias.cairo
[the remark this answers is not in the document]
aymericdelab: @milancermak I'm hesitating to use type ID for resource_id here because I don't think resource_id is actually an ID but rather a type (not more than 255 resource types so u8 works here). So i'm thinking it might be better to rename as resource_type and keep u8 here.

```
fn has_resource(self: Realm, resource_id: u8) -> bool {
        let mut resource_ids: Span<u8> = unpack_resource_ids(
            self.resource_ids_packed, self.resource_ids_count
        );
```

src/components/army.cairo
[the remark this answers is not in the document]
milancermak: Whoa, that sounds like a great idea! 🧠 

I'll resolve this convo, but let's revisit this once we tackle armies and combat.

src/alias.cairo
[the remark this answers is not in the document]
milancermak: Sound good! I agree with all.

src/systems/settling.cairo
```diff
@@ -0,0 +1,289 @@
...
+                    resource_ids_packed: realm_data.resource_ids_packed,
+                    resource_ids_count: realm_data.resource_ids_count,
+                    cities: realm_data.cities,
+                    harbors: realm_data.harbors,
+                    rivers: realm_data.rivers,
+                    regions: realm_data.regions,
+                    wonder: realm_data.wonder,
+                    order: realm_data.order,
+                    }, Owner {
+                    address: owner.address,
```
milancermak, line 48: Follow-up on the discussion above, the address here should not be set to `owner.address` but rather the tx_info account address.

src/systems/settling.cairo
```diff
@@ -0,0 +1,289 @@
...
+            let resource_query: Query = (realm_id, resource_id_felt).into();
+            commands::<Resource>::set_entity(
+                resource_query,
+                (Resource { id: resource_id, balance: config.base_resources_per_day,  })
+            );
+            index += 1;
+        }
+
+        // transfer Realm ERC721 to world contract
+        erc721.transfer_from(owner.address, world_address, realm_id, );
```
milancermak, line 74: Similar here, `owner.address` won't work.

src/systems/settling.cairo
```diff
@@ -0,0 +1,289 @@
...
+            realm_id,
+        );
+
+        // delete entity
+        let world = IWorldDispatcher {contract_address: world_address};
+        world.delete_entity('Owner'.into(), realm_query);
+        world.delete_entity('Position'.into(), realm_query);
+        world.delete_entity('Realm'.into(), realm_query);
+        world.delete_entity('Age'.into(), realm_query);
+        // TODO: should delete resources ?
```
milancermak, line 117: I guess it's not necessary because the resource entities will get re-set on settling again.

milancermak (commented):
Left just a couple of small comments, otherwise it's all looking good.

src/systems/settling.cairo
aymericdelab: not sure why that wouldn't work though? 
the idea is to do this:
```
let owner = commands::<Owner>::entity((token, realm_id.into()).into());
let caller = starknet::get_tx_info().unbox().account_contract_address;
assert(owner.address == caller, 'Only owner can settle');
```
So owner.address and tx_info().account_contract_address should be the same thing here. I had to comment it out though because it will only be available in next release

milancermak: But there's no Owner entity yet, is there? The entity has not been set, so `commands::<Owner>::entity((token, realm_id.into()).into())` returns an empty/zero owner if I'm not mistaken.

aymericdelab: Owner is set during mint 
```
let token: felt252 = config.realm_l2_contract.into();
let owner = commands::<Owner>::entity((token, realm_id.into()).into());
```
this retrieves entity defined by ERC721

ponderingdemocritus:
Lets ship thiss

src/systems/settling.cairo
milancermak: Resolving to unlock merging

src/systems/settling.cairo
milancermak: Resolving to unlock merging

PR #12, opened 2023-05-15
fix: remove derive magic

aymericdelab:
quick fix because dojo does not assume the traits anymore

### DIFF
--- a/src/components/age.cairo
+++ b/src/components/age.cairo
@@ -1,4 +1,4 @@
-#[derive(Component)]
+#[derive(Component, Copy, Drop, Serde)]
 struct Age {
     born_at: u64, 
 }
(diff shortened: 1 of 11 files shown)


src/components/age.cairo
```diff
@@ -25,20 +25,16 @@ impl AgeImpl of AgeTrait {
...
-        born_at: 100,
-    };
+    let age = Age { born_at: 100,  };
     let age_difference = age.get_age_difference(200);
     assert(age_difference == 100, 'Age difference should be 100');
 }
 
 #[test]
 #[available_gas(30000000)]
 fn get_current_age() {
```
milancermak, line 35: Unrelated to this PR, but the test function names should be prefixed with `test_`

milancermak (approved):
Cool stuff. Left just a comment about style, not related to the PR itself. Ship it!

PR #13, opened 2023-05-23
fix: new dojo auth system

aymericdelab:
new auth system implemented in dojo. 

in the tests spawn_test_world now needs an array of routes to give systems the necessary authorizations. 
a route is comprised of 
1. system that needs to write
2. role name
3. component it writes to


ponderingdemocritus:
LGTM - I didn't run tests, but happy to merge if you are

PR #14, opened 2023-05-25
feat: pseudo code version

aymericdelab:
this is a first version of the code for the p2p trading. 

The **maker** = the one creating the order
The **taker** = the one accepting the order

for now you have 3 systems:
(summarized by chatgpt)

- **CreateOrder**: the maker can create an order by specifying the amounts of resources its willing to swap. An array of entities that will be used in the maker caravan needs to be specified as well. If not enough units for the total weight, free units are used.
- **AcceptOrder**: a taker can accept the order by also specifying the units that will be used in the caravans. If not enough units for the total weight, free units are used. When taker accepts the order, both the taker and maker caravans can start travelling. 
- **ClaimOrder**: both the maker and taker can call this system in order to verify that the caravans arrived at the destination, if that's the case the swapped resources can be added to their balance.

### CreateOrder
- The function takes several input parameters, including the maker's ID, resource types and quantities, expiration time, and a list of entity IDs representing a caravan.
- The function calculates the total weight of the maker's resources based on the provided resource type and quantity.
- It generates a unique order ID using the commands::uuid() function.
- The function iterates over the caravan entities and determines if they can carry the remaining weight of the resources. If a caravan unit can carry the weight, it is attached to the order entity.
- If the capacity limit is reached, additional free transport units are created and attached to the order entity.
- The order entity is created and filled with relevant information, including the maker's details, resource type and quantity, caravan count, expiration time, and order status.
- The function asserts that the maker has enough resources in their realm balance.
- The maker's realm balance is updated by subtracting the resource quantity of the order from the current balance.

### AcceptOrder
- The function takes several input parameters, including the taker's ID, the order ID, and a list of entity IDs representing a taker caravan.
- The function performs various assertions to validate the inputs, such as checking if the taker is a realm entity and if the caller is the owner of the taker realm.
- The function calculates the current block timestamp using the get_block_timestamp() function.
- The function iterates over the taker caravan entities and determines if they can carry the remaining weight of the resources. If a caravan unit can carry the weight, it is attached to the order entity.
- If the capacity limit is reached, additional free transport units are created and attached to the order entity.
- The function calculates the average speed of the taker caravan units based on their total speed and quantity.
- It calculates the arrival time of the taker at the maker's position and updates the taker caravan units' arrival times accordingly.
- It calculates the arrival time of the maker at their own position based on the maker caravan units' speed and quantity.
- The function updates the maker caravan units' arrival times accordingly.
- It checks if the taker has enough resources in their realm balance to fulfill the order and updates the balance by subtracting the taker resource quantity.
- The function updates the order entity with the maker's details, resource type and quantity, arrival times, taker details, and order status as "ACCEPTED".

### ClaimOrder
- The function takes two input parameters: entity_id and order_id.
- The function performs various assertions to validate the inputs, such as checking if the caller is the owner of entity_id and if entity_id is a realm entity.
- It retrieves the order details from the Order entity using the order_id.
- The function retrieves the current block timestamp using the get_block_timestamp() function.
- If the caller is the maker of the order, it verifies that the taker's resources have arrived at the maker's realm by checking the taker's arrival time.
- It adds the taker's resources to the maker's resource balance by updating the balance of the specified resource type.
- It sets the taker's resource quantity to 0 in the order entity to prevent the taker from claiming it again.
- If the caller is the taker, it verifies that the maker's resources have arrived at the taker's realm by checking the maker's arrival time.
- It adds the maker's resources to the taker's resource balance by updating the balance of the specified resource type.
- It sets the maker's resource quantity to 0 in the order entity to prevent the maker from claiming it again.


aymericdelab:
simple trading system for fungible entities = entities that can be defined by only their entity_type (resource, food, coins) and their quantity.
Feats:
- create and add caravans independently of the order creation
- possibility to trade between 2 entities on the same position
- can trade list of fungible entities against list of fungible entities

src/components/entities.cairo
```diff
@@ -0,0 +1,19 @@
+use eternum::alias::ID;
+
+#[derive(Component, Copy, Drop, Serde)]
+struct Entities {
```
ponderingdemocritus, line 4: this is confusing I feel - we need a different naming pattern

PR #15, opened 2023-06-09
fix: refactor due to dojo changes

aymericdelab:
some refactoring in milestone 1 due to breaking changes in dojo


aymericdelab:
can we get this merged ? @ponderingdemocritus

PR #36, opened 2023-07-28
refactor: env variables + prod/dev

aymericdelab:
### client
set .env.dev and prod to handle env variables in client

### scripts
+ added ./scripts/set_config.sh with prod keyword to wait 3 seconds between calls in order to not overload remote node.
+ added prod option when calling `source ./scripts/env_variables.sh in order to set env variable


r0man1337 (approved):
LGTM

PR #84, opened 2023-08-22
fix: updated torii and katana instance URLs

r0man1337:
Since rinnengan instance no longer works, I installed katana and torii instances on my VPS and changed env variables to it.


aymericdelab:
Can we still keep the madara env ? So put what was in the .env.production inside .env.madara or something

PR #85, opened 2023-08-22
Contracts/update dojo

aymericdelab:
@ponderingdemocritus only change i made to the contracts other than dojo-updates, is the use of a new component "OrderResource". 
Previously I was using "Resource" component (used for realms balances), to also represent the resources in an order. 
But for the former I need it indexed with "entity_id, resource_type" and fields "balance" while for the latter i need it indexed with "order_id, fungible_entities_id, index" and with fields "resource_type, balance".

i think it makes more sense as it utilmately does not represent the same thing in the world.

so now we've got:
```
struct Resource {
    // This is compound key of entity_id and resource_type
    #[key]
    entity_id: u128,
    #[key]
    resource_type: u8,
    balance: u128,
}
```
and 
```
#[derive(Component, Copy, Drop, Serde, SerdeLen)]
struct OrderResource {
    #[key]
    order_id: u128,
    #[key]
    fungible_entities_id: u128,
    #[key]
    index: usize,
    resource_type: u8,
    balance: u128,
}
```


r0man1337 (commented):
lgtm

r0man1337 (approved):

Issue #86, opened 2023-08-22
Update tests


aymericdelab:
There are already a bunch of tests done on the main systems, these were working a couple months ago before there were some breaking changes in Dojo so we need to refactor.
It would also be interesting that you review the way the tests were done and see if we can improve.
Roll Your Own is the other big project on dojo (https://github.com/cartridge-gg/rollyourown). Would be interesting to take a look at how they do their testing.

**Current way of doing testing for a system:**
1. call spawn_test_world_without_init() from src/utils/testing.cairo, which will deploy a world and register the class hash of all systems and component, but it wont initialise it, this is a trick we're using in testing, because when you init your world in dojo, it means you need to handle permissions between systems and component as well, which we dont want to do for testing.
2. set necessary configs: some values from the world (like the speed of a specific transport unit), are set in specific config components so that we can have modify these to balance the game. 

## TODO:

- [x] refactor testing.cairo

**Systems for which we need to update current tests:**

### Labor
- [x] build_labor
- [x] harvest_labor

### Caravan
- [x] create_caravan
- [x] create_free_transport_unit
- [x] utils

### Order (trading)
- [x] make_fungible_order
- [x] take_fungible_order
- [x] claim_fungible_order
- [x] attach_caravan

PR #89, opened 2023-08-23
Update create_caravan test

credence0x:
PR for #86 
- update create_caravan test


ponderingdemocritus:
Going to merge - but we need to fix actions so we confirm this passes

PR #94, opened 2023-08-24
refactor tests

credence0x:
PR for #86 

- created a common `setup_eternum` function to reduce code duplication  across tests
- Updated tests for caravan utils


credence0x:
The `GetQuantity` system tests were removed entirely because we plan to remove the system

PR #105, opened 2023-08-27
update take_fungible_order system test

credence0x:
PR for #86 
- update take_fungible_order system test

### DIFF
--- a/contracts/src/systems/order/take_fungible_order.cairo
+++ b/contracts/src/systems/order/take_fungible_order.cairo
@@ -459,6 +459,11 @@ mod tests {
         set!(world, (Capacity { weight_gram: 10_000, entity_id: caravan_id.into()}));
         set!(world, (Movable { sec_per_km: 10, blocked: false, entity_id: caravan_id.into()}));
 
+        // attach caravan to the maker order
+        let caravan_key_arr = array![maker_order_id.into(), maker_id.into()];
+        let caravan_key = poseidon_hash_span(caravan_key_arr.span());
+        set!(world, (Caravan { caravan_id: caravan_id.into(), entity_id: caravan_key }));
+
 
         // attach caravan to the taker order
         let caravan_key_arr = array![taker_order_id.into(), taker_id.into()];
@@ -504,8 +509,8 @@ mod tests {
 
         // verify position of maker order
         let maker_order_position = get!(world, maker_order_id, Position);
-        assert(maker_order_position.x == 45, 'position x should be 45');
-        assert(maker_order_position.y == 50, 'position y should be 50');
+        assert(maker_order_position.x == 60, 'position x should be 60');
+        assert(maker_order_position.y == 70, 'position y should be 70');
 
 
         // verify position of taker order
--- a/contracts/src/systems/order/take_fungible_order.cairo
+++ b/contracts/src/systems/order/take_fungible_order.cairo
@@ -451,24 +451,31 @@ mod tests {
             taker_order_id
         ) = setup(true);
 
-        // create a caravan owned by the taker
-        let caravan_id = 20_u64;
-        let caravan_id_felt: felt252 = caravan_id.into();
-        set!(world, (Owner { address: contract_address_const::<'taker'>(), entity_id: caravan_id.into()}));
-        set!(world, (Position { x: 45, y: 50, entity_id: caravan_id.into()}));
-        set!(world, (Capacity { weight_gram: 10_000, entity_id: caravan_id.into()}));
-        set!(world, (Movable { sec_per_km: 10, blocked: false, entity_id: caravan_id.into()}));
+        // create a caravan owned by the maker
+        let maker_caravan_id = 20_u64;
+        let maker_caravan_id_felt: felt252 = maker_caravan_id.into();
 
+        set!(world, (Owner { address: contract_address_const::<'maker'>(), entity_id: maker_caravan_id.into()}));
+        set!(world, (Position { x: 100_000, y: 200_000, entity_id: maker_caravan_id.into()}));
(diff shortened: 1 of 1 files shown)


contracts/src/systems/order/take_fungible_order.cairo
```diff
@@ -232,361 +232,287 @@ mod TakeFungibleOrder {
...
+    fn test_take_trade_with_caravan() {
+        
+        let (
+            world,
+            maker_id,
+            taker_id, 
+            trade_id, 
+            maker_order_id, 
+            taker_order_id
+        ) = setup();
```
aymericdelab, line 451: If you want to test with caravans, you need the maker to create the trade struct with field:  taker_needs_caravan: true.
If you don't this allows the taker to choose not to use a caravan.

So here in your test you should change taker_needs_caravan to true. And attach a caravan for the taker as well

credence0x: Updated

contracts/src/systems/order/take_fungible_order.cairo
```diff
@@ -232,361 +233,292 @@ mod TakeFungibleOrder {
...
+
+
+        // status should be accepted
+        let status = get!(world, trade_id, Status);
+        assert(status.value == 1, 'status should be accepted');
+
+        
+        // verify arrival time of maker order
+        let maker_arrival_time = get!(world, maker_order_id, ArrivalTime);
+        assert(maker_arrival_time.arrives_at == 0, 'arrival time should be 0');
```
aymericdelab, line 502: arrival time should be in the future if both use caravans

contracts/src/systems/order/take_fungible_order.cairo
```diff
@@ -232,361 +233,292 @@ mod TakeFungibleOrder {
...
+ 
+
+        // verify arrival time of taker order
+        let taker_arrival_time = get!(world, taker_order_id, ArrivalTime);
+        assert(taker_arrival_time.arrives_at == 0, 'arrival time should be 0');
```
aymericdelab, line 507: arrival time should be in the future if both uses caravans

contracts/src/systems/order/take_fungible_order.cairo
```diff
@@ -232,361 +233,292 @@ mod TakeFungibleOrder {
...
+        let caravan_id_felt: felt252 = caravan_id.into();
+        set!(world, (Owner { address: contract_address_const::<'taker'>(), entity_id: caravan_id.into()}));
+        set!(world, (Position { x: 45, y: 50, entity_id: caravan_id.into()}));
+        set!(world, (Capacity { weight_gram: 10_000, entity_id: caravan_id.into()}));
+        set!(world, (Movable { sec_per_km: 10, blocked: false, entity_id: caravan_id.into()}));
+
+        // attach caravan to the maker order
+        let caravan_key_arr = array![maker_order_id.into(), maker_id.into()];
+        let caravan_key = poseidon_hash_span(caravan_key_arr.span());
+        set!(world, (Caravan { caravan_id: caravan_id.into(), entity_id: caravan_key }));
```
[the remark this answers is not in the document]
credence0x: Correct thanks. Fixed

contracts/src/systems/order/take_fungible_order.cairo
```diff
@@ -232,361 +233,301 @@ mod TakeFungibleOrder {
...
+            maker_order_id, 
+            taker_order_id
+        ) = setup(true);
+
+        // create a caravan owned by the maker
+        let maker_caravan_id = 20_u64;
+        let maker_caravan_id_felt: felt252 = maker_caravan_id.into();
+
+        set!(world, (Owner { address: contract_address_const::<'maker'>(), entity_id: maker_caravan_id.into()}));
+        set!(world, (Position { x: 100_000, y: 200_000, entity_id: maker_caravan_id.into()}));
```
aymericdelab, line 459: you should set the position of the maker caravan on the maker position, because that is the only way to attach the caravan.
Same for the taker caravan position which should be on the same position as the taker.

contracts/src/systems/order/take_fungible_order.cairo
```diff
@@ -232,361 +233,301 @@ mod TakeFungibleOrder {
...
+
+        // verify arrival time and position of taker order
+        let taker_order_arrival_time = get!(world, taker_order_id, ArrivalTime);
+        let taker_order_position = get!(world, taker_order_id, Position);
+        assert(taker_order_arrival_time.arrives_at == 900, 'arrival time should be 900');
+        assert(taker_order_position.x == 45, 'position x should be 45');
+        assert(taker_order_position.y == 50, 'position y should be 50');
+        // verify arrival time of taker caravan
+        let taker_caravan_arrival_time = get!(world, taker_caravan_id, ArrivalTime);
+        assert(taker_caravan_arrival_time.arrives_at == (900 * 2), 'arrival time should be 1800');
```
aymericdelab, line 527: if both realms have a caravan with the same speed, the taker and maker caravans should arrive at the same time

aymericdelab (changes requested):

contracts/src/systems/order/take_fungible_order.cairo
credence0x: Got it. Fixed. Updated the taker and maker positions to large values also because the distance between small values like  `Position {45, 60}` and `Position {60,70}` would be 0, making travel time 0

contracts/src/systems/order/take_fungible_order.cairo
credence0x: Fixed

contracts/src/systems/order/take_fungible_order.cairo
aymericdelab: ok

aymericdelab (approved):

Issue #66, opened 2023-08-17
Notifications menu


Fomentador:
Eternum needs a notification menu. The game is feature rich, meaning there'll be a lot going on at all times, and it's played in a real time, competitive multiplayer environment, meaning losing track of any relevant information may result in significant damage to your Realms. Thus, gameplay-relevant information must always be highlighted and delivered to the player in a pronounced, prominent way, and this way is the notifications menu.

- The notifications menu is player-based, and will deliver only the information relevant for that specific player's empire. It'll have two versions: one with notifications pertaining only a specific Realm (standard setting in the Realms map), and one with notifications pertaining multiples Realms (standard setting showing all Realms in your wallet, in the world map). 

- Additionally, a configurations menu should allow the player to customize whichever version of notifications they want to see where, e.g. if the player wants to see notifications for all Realms when in the Realms map or even if they want to pick and choose from their Realms. Ideally that could manifest as a toggle button to show notifications for all Realms even in the Realms map, with the addition of an option to checkmark specific Realms to be shown in notifications. The configurations menu should also have a toggle for dropdowns menus in notifications (show them opened or the default, closed).

- Notifications should not only be closable/dismissible, but new notifications should also be highlighted in the list to indicate they're new. The highlight can be done several ways: background, borders, colours, glow, among others. A new notifications loses its highlight once the player passes the mouse over it.

- Notifications can also be pinned and unpinned.

- Notifications' heading will redirect the player to the respective UI menu for that information.

A first mock-up was already made couple months ago:

<img width="466" alt="notifications-mockup" src="https://github.com/BibliothecaDAO/eternum/assets/110478008/044e78dd-533f-4336-8e88-97bfd8abdd1c">

Below is the list of notifications needed, with the caveat that as more game systems are implemented into the game (Banking, Crafting, etc.) the list of notifications needed will also expand. The list contains the information for both notification menu types: single Realm and multiple Realms.

- [ ] Offer accepted.
For every time a standing offer is accepted. Notification heading redirects to the Offers menu.

- Single Realm version:
Contains simply the text "Offer accepted" and nothing else. A dropdown menu will show resources sent, resources received (both individually), Merchants in Caravan, partner Realm, and real time distance remaining in hours.

- Multiple Realms versions:
Same as single Realm, with the addition of the Realm name in the heading by the side of "Offer accepted".

- [ ] Offer proposed. 
For every time someone proposes an offer directly to your Realm. Notification heading redirects to the Offers menu.

[...]

PR #109, opened 2023-08-28
feat: entity updates + notifications

### LINKED ISSUE #106: [client] Component subscription
We now have entities/components subscriptions in Torii. 
In the client the idea would be to be able to subscribe to the following components:

https://github.com/BibliothecaDAO/eternum/blob/main/client/src/dojo/contractComponents.ts
(not the config ones)

And update the internal tables using setComponent every time there is a change.

I think best would be to create a componentSubscription.ts file in dojo folder then instantiate the subscription in setup.ts.

aymericdelab:
fix #106 + notifications from #66

- I'm using graphql-ws to subscribe to torii ws.

- I use graphql query entityUpdated to be notified when there is an entity update, but unfortunately we don't have the component values with it yet (will be implemented soon). So I'm temporarily forced to make a second query to torii to get the entity component values.

- I create an Observable "entityUpdates" that I can access from anywhere in the app.

**UseNotifications custom hook**:
 1. retrieves the list of entityUpdates from Torii and filters them + gives them the right EventType
 2. checks if the labor can be harvested or the incoming resources are claimable
   
**Notification Types**
1. Accept/Create/Cancel Offer
2. Incoming resources claimable
3. Labor ready to be harvested

**Missing features from #66 that should be added later**
- historical notifications (ex: your offers accepted before page refresh)
- notifications management (filtering)
- notification interaction (click on harvest notification and it harvests the labor, click on resources claimable and it should claim them, ...)


aymericdelab:
@r0man1337 could take a look at : 
1. how to deal with overflow when there are too many notifications
2. better layout for the claimable order notifications

r0man1337 (approved):

Issue #111, opened 2023-08-29
[contract] Cancel Order system


aymericdelab:
Currently we are using ChangeOrderStatus to cancel an order by just changing it's status to "Cancel" (status = 2).

We should remove the ChangeOrderStatus system and replace it by CancelOrder. That system would:

- make sure caller is maker
- make sure the order is open (status 0) and not accepted (status 1)
- change status from 0 to 2
- put the maker resources from OrderResource back in maker balances
- remove the caravan from the order and unblocking it

**+ add necessary testing**

aymericdelab:
@credence0x can you take a look at this?

credence0x:
Sure

PR #112, opened 2023-08-29
Cancel fungible order

credence0x:
PR for #111 
- created helpers for orders
- created and tested cancel order system
- deleted change order system


contracts/src/systems/order/helpers/order.cairo
```diff
@@ -0,0 +1,154 @@
...
+
+    fn attach(world: IWorldDispatcher, caravan_id: ID, order_id: ID, owner_id: ID) {
+                
+        let caravan = get(world, order_id, owner_id);
+        assert(caravan.caravan_id == 0, 'caravan already attached');
+
+        let caravan_movable = get!(world, caravan.caravan_id, Movable);        
+        assert(caravan_movable.blocked == false, 'caravan already blocked');
+
+        //q: should we check if caravan is attached to another order?
```
aymericdelab, line 132: if caravan is already attached, then it will be blocked (assert on line above)

contracts/src/systems/order/cancel_fungible_order.cairo
```diff
@@ -0,0 +1,258 @@
...
+    use eternum::systems::order::helpers::order;
+
+    use box::BoxTrait;
+    use traits::Into;
+
+    use dojo::world::Context;
+
+    fn execute(ctx: Context, trade_id: u128) {
+
+        let (trade_meta, trade_status) = get!(ctx.world, trade_id, (Trade, Status));
```
aymericdelab, line 15: i like that you use trade_meta instead of meta, can you check if there are other places where we use meta or status instead of trade_meta and trade_status and refactor?

aymericdelab (approved):
looks good to me, i'll add a PR to modify the client to incorporate CancelFungibleOrder

PR #117, opened 2023-09-04
update world address, add burner create, add in address

ponderingdemocritus:
- Adds burner wallet creation
- adds script for gen components rather than local file
- updates some deterministic vars


aymericdelab:
@ponderingdemocritus getting error:
```
sequencer.ts:235 Uncaught (in promise) GatewayError: max_fee must be bigger than 0.
0 >= 0
```
when deploying new burner

aymericdelab (approved):

PR #126, opened 2023-09-07
fix: change world contract

aymericdelab:


aymericdelab:
@ponderingdemocritus lgtm

r0man1337 (approved):

PR #132, opened 2023-09-09
Create hyperstructure feature [contract]

credence0x:
PR for #134 
- Created necessary components and systems for hyperstructures
- Created system to transfer resources between entities
- Created system to allow movable entities travel
- Created `RoadConfig` component  to allow for configuration flexibility


contracts/src/systems/craft.cairo
```diff
@@ -0,0 +1,71 @@
...
+        loop {
+            if i == materials.len() {
+                break;
+            }
+
+            let material: StructureMaterial = *materials[i];
+            assert(material.amount != 0, 'material amount is 0');
+
+
+            let owner = get!(ctx.world, caller_id, Owner);
```
aymericdelab, line 31: why not checkiing that outside the loop ?

aymericdelab:
very nice implementation, I like the way you define these recipes as a hash of the necessary materials.

contracts/src/components.cairo
```diff
@@ -11,4 +11,5 @@ mod quantity;
 mod metadata;
 mod caravan;
 mod trade;
-mod road;
\ No newline at end of file
+mod road;
+mod structure;
```
aymericdelab, line 15: i think the correct term for these is hyperstructures, can you change that everywhere in the code ?

aymericdelab: https://ebrary.net/127150/management/visions_city_future
After searching in chat history, i believe this is the inspiration for hyperstructures if you want to read a bit about it

contracts/src/systems/craft.cairo
```diff
@@ -0,0 +1,71 @@
+#[system]
+mod Craft {
```
aymericdelab, line 2: i wouldnt call this crafting, crafting is an entirely different part of the game where we can create different materials from resources.

contracts/src/systems/craft.cairo
```diff
@@ -0,0 +1,71 @@
...
+                }
+            ));
+
+            Serde::serialize(@material, ref materials_felt_array);
+
+            i+=1;
+        };
+        
+
+        let structure_composition_hash: felt252 = poseidon_hash_span(materials_felt_array.span());
```
aymericdelab, line 51: im thinking there might be a mixup between crafting and hyperstructures. I think what you did might be closer to crafting and we could keep the code for that

contracts/src/systems/craft.cairo
```diff
@@ -0,0 +1,71 @@
...
+    use core::serde::Serde;
+
+
+
+    fn execute(ctx: Context, caller_id: ID, materials: Span<StructureMaterial>) -> ID {
+
+        let mut materials_felt_array: Array<felt252> = array![];
+        let mut i = 0;
+        loop {
+            if i == materials.len() {
```
ponderingdemocritus, line 23: match and pop_front() is more efficient

contracts/src/components/hyperstructure.cairo
```diff
@@ -0,0 +1,20 @@
...
+    entity_id: u128,
+    created_at: u64,
+    completed_at: u64,
+    resource_count: usize
+}
+
+
+
+#[derive(Component, Copy, Drop, Serde)]
+struct HyperStructureResource {
```
ponderingdemocritus, line 13: do we need this? can we not use the other resource structure?

contracts/src/components/hyperstructure.cairo
```diff
@@ -0,0 +1,20 @@
+#[derive(Component, Copy, Drop, Serde, SerdeLen)]
+struct HyperStructure {
+    #[key]
+    entity_id: u128,
+    created_at: u64,
+    completed_at: u64,
+    resource_count: usize
```
ponderingdemocritus, line 7: let's add a 'hyperstructure_id' here too to define the type

contracts/src/systems/hyperstructure/transport_to_hyperstructure.cairo
```diff
@@ -0,0 +1,75 @@
+#[system]
+mod TransportToHyperStructure {
```
ponderingdemocritus, line 2: I think this system could be generalised with the transport system, as it is only shipping from a->b

contracts/src/systems/hyperstructure/initialize_hyperstructure.cairo
```diff
@@ -0,0 +1,67 @@
...
+
+    fn execute(ctx: Context, resources: Span<ReqiuredResource>, coord: Coord) -> ID {   
+
+        // todo@credence: check permissions
+        // todo@credence: maybe ensure resources aren't duplicated
+
+        let hyperstructure_id: ID = ctx.world.uuid().into();
+        let mut index = 0;
+        loop {
+            if index == resources.len() {
```
ponderingdemocritus, line 27: pop_front()

contracts/src/systems/hyperstructure/initialize_hyperstructure.cairo
```diff
@@ -0,0 +1,67 @@
...
+    use core::traits::Into;
+    use core::array::{SpanTrait};
+
+    #[derive(Copy, Drop, Serde)]
+    struct ReqiuredResource {
+        resource_type: u8,
+        amount: usize
+    }
+
+    fn execute(ctx: Context, resources: Span<ReqiuredResource>, coord: Coord) -> ID {
```
ponderingdemocritus, line 19: the resource required should as a constant value somewhere, no passed in. We should just pass a 'hyperstructure_id' which will denote what the cost will be.

We will then need to track the balance of the structure built

contracts/src/utils/caravan.cairo
```diff
@@ -0,0 +1,198 @@
...
+    caravan_attachment.cargo_id = 0;
+
+    set!(world, (caravan_attachment, cargo));
+
+    cargo
+}
+
+
+
+mod cargo {
```
ponderingdemocritus, line 109: should this be a trait rather than mod?

contracts/src/systems/config/hyperstructure_config.cairo
```diff
@@ -0,0 +1,64 @@
...
+    use eternum::components::position::{Position, Coord};
+
+    use dojo::world::Context;
+
+    use core::poseidon::poseidon_hash_span;
+    use core::traits::Into;
+    use core::array::{SpanTrait};
+
+    #[derive(Copy, Drop, Serde)]
+    struct ReqiuredResource {
```
aymericdelab, line 14: spelling

contracts/src/systems/hyperstructure/initialize_hyperstructure.cairo
```diff
@@ -0,0 +1,26 @@
+#[system]
+mod InitializeHyperStructure {
+    use eternum::alias::ID;
+    use eternum::components::hyperstructure::HyperStructure;
+
+    use dojo::world::Context;
+
+    fn execute(ctx: Context, entity_id:ID, hyperstructure_id: ID) {
```
aymericdelab, line 8: my initial thought was that initializing the hyperstructure would cost resources as well, what do you think @ponderingdemocritus

contracts/src/systems/hyperstructure/transport_to_hyperstructure.cairo
aymericdelab: so the flow for hyperstructures using 1-way Travels would be:

1. Travel Realm -> Hyperstruct (then wait for the caravan to arrive)
2. Offload Caravan
3. Travel back to Realm 

Using 2-way Travels you could have:
1. Travel back and forth from Realm -> Hyperstructure -> Realm.
2. Offload Caravan when we know the caravan is on the way back from the Hyperstructure

Option 2 is similar to what is done in trading (albeit not generalizable to other cases yet)

Would be interesting to discuss both approaches together. @credence0x @ponderingdemocritus

contracts/src/systems/hyperstructure/transport_to_hyperstructure.cairo
```diff
@@ -0,0 +1,75 @@
...
+
+            index+=1;
+        };
+
+
+        // travel from Realm --> Hyperstructure 
+        let entity_position = get!(ctx.world, entity_id, Position);
+        let hyperstructure_position = get!(ctx.world, hyperstructure.entity_id, Position);
+
+        caravan::onboard(ctx.world, caravan_id, cargo);
```
aymericdelab, line 70: I'm thinking, does it make more sense to create a TransferResources entrypoint ? Onboarding the resources is the same as transfering resources to another entity (caravan here). TransferResources would check:
1. are you on the same position?
2. check if you have enough capactity, only if you are MOVABLE. (Because lands, hyperstrucutres and these non movable entities should have "unlimited" capacity).

Onloading and offloading would basically be just TransferResources.

contracts/src/systems/caravan/move_caravan.cairo
```diff
@@ -0,0 +1,64 @@
+#[system]
```
aymericdelab, line 1: this is exactily what we need, one remark though is that we should not link this to the caravans. Caravan is a way to put multiple movable entities in the same entity and move them together. 
But Travel should be for any movable entity, not just caravans. What I would suggest is to:
1. rename this to mod Travel
2. remove any notion of caravan in the code. It just needs to be anything that is movable and has a position.
3. create a folder travel and put it in there

contracts/src/systems/config/hyperstructure_config.cairo
```diff
@@ -0,0 +1,61 @@
...
+    use core::traits::Into;
+    use core::array::{SpanTrait};
+
+    #[derive(Copy, Drop, Serde)]
+    struct RequiredResource {
+        resource_type: u8,
+        amount: usize
+    }
+
+    fn execute(ctx: Context, mut resources: Span<RequiredResource>, coord: Coord) -> ID {
```
aymericdelab, line 19: maybe add a TODO to not forget to put that in ./set_config when we know what resources should be needed for each hyperstruture.

contracts/src/systems/hyperstructure/initialize_hyperstructure.cairo
aymericdelab: same comment as above

contracts/src/systems/resources/transfer_resources.cairo
```diff
@@ -0,0 +1,74 @@
...
+        let sending_entity_position = get!(ctx.world, sending_entity_id, Position);
+        let receiving_entity_position = get!(ctx.world, receiving_entity_id, Position);
+
+        assert(receiving_entity_position.x !=  0, 'entity position mismatch');
+        assert(receiving_entity_position.y != 0, 'entity position mismatch');
+
+        assert(receiving_entity_position.x == sending_entity_position.x, 'entity position mismatch');
+        assert(receiving_entity_position.y == sending_entity_position.y, 'entity position mismatch');
+       
+        // get receiving entity's total capacity
```
aymericdelab, line 29: this is good, one remark:
if the receiving entity does not have a capacity and is not movable, then it can receive any amount of resources.
The logic is that if no capacity/movable, then it must be something like a Realm or Hyperstructure. On which you can put unlimited amount of resources.

I think this would work for now, let's see in the future if there are flaws in this approach.

contracts/src/components/hyperstructure.cairo
aymericdelab: Issue still seems to be the same as with OrderResources.

We have a Resource component where entity_id, resource_type are the keys, and we have only amount as value.

Here we nee

contracts/src/components/hyperstructure.cairo
aymericdelab: @credence0x as loaf noted, in the future we migth want to have different types of hyperstructures. That can give different bonuses if completed. 
So adding a hyperstructure_type field here would help us differentiate between these hyperstructures. 

Let's say the ones we create for each order are going to be the first hyperstrucutres. And let's give them hyperstructure_type 1.

contracts/src/systems/config/hyperstructure_config.cairo
aymericdelab: would add hyperstrucutre_type here as well then

contracts/src/systems/caravan/move_caravan.cairo
aymericdelab: i think you're mistaking. The Movable component means anything that can move by itself, not movable goods. So like a horse is movable. An army is movable. Gold is not movable. But it can be moved by a movable entity. 
I agree that "Movable" might be confusing though.
As for only caravans being able to move things. I think that would go against ECS. In our implementation, any entity that has a Capacity component and a Movable component, should be able to move resources around.

contracts/src/systems/resources/transfer_resources.cairo
aymericdelab: ok i see, that's ok

contracts/src/systems/config/structure_config.cairo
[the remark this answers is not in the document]
aymericdelab: i think the upside of having devs know how much each thing cost in the game by only looking at the code, is less than the upside of having these numbers changeable as we are testing rn and not sure what is best.

Also, possible feature would be to have the DAO decide on changing these numbers as well

contracts/src/systems/caravan/move_caravan.cairo
credence0x: You're right

contracts/src/systems/craft.cairo
credence0x: Didn't want to make the PR too bloated so i didn't include it. It's here just incase

contracts/scripts/set_config.sh
```diff
@@ -80,6 +87,15 @@ commands+=(
     "sozo execute --world "$world" SetLaborCostAmount --account-address $DOJO_ACCOUNT_ADDRESS --calldata 255,3,10"
 )
 
+
+commands+=(
+    # Define hyperstructure
+    # hyperstructure type 1
+    # resource 1,000 stone and 1,000 shekels
+    # hyperstructure coordinate x: 800, y : 200 
+    "sozo execute --world $world DefineHyperStructure --account-address $DOJO_ACCOUNT_ADDRESS --calldata 1, 2,2,1000,3,1000, 800,200"
```
credence0x, line 96: @aymericdelab Used arbitrary values. You can  reply with correct values if necessary

contracts/src/components/road.cairo
```diff
@@ -42,18 +32,6 @@ impl RoadImpl of RoadTrait {
         road
     }
 
-
-    #[inline(always)]
-    fn speed_boost(self: Road) -> u64 {
```
credence0x, line 47: This was removed because `RoadConfig` was introduced

contracts/src/components/road.cairo
```diff
@@ -42,18 +32,6 @@ impl RoadImpl of RoadTrait {
...
-        2   
-    }
-
-
-    #[inline(always)]
-    fn travel(ref self: Road, world: IWorldDispatcher) {
```
credence0x, line 53: This was removed because i think after we clean out the codebase, `usage_count ` would only need to be updated in the travel system or a related one.

aymericdelab (approved):

PR #144, opened 2023-09-15
feat: draggable popups

r0man1337:
- Labor, Settings and Create offer popups are now draggable.
- After dragging popup position is saved to localStorage, and this position will be used as default after reopening the popup between game sessions.
- Clicking on popup now moves it to top level z-index
- Last opened popup appears on top level z-index


ponderingdemocritus:
This is awesome! LGTM

PR #147, opened 2023-09-18
feat: add auction component + get_zone in position

aymericdelab:


aymericdelab:
Added simple UI integration of the labor prices.

I'm showing labor info on 2 different places:

1. Under the Realm Status: zone number + current labor price for one extra unit of labor (2 hours)
2. When building labor, the resource price is multiplied by average labor price for selected number of units, which is shown below. + I'm showing what will be the next labor price after building all this labor.

Screenshot 2023-09-23 at 20 50 23

aymericdelab:
[WARNING]
sometimes vrgda needs more steps than allowed in default max number of steps per tx. So you need to launch katana with:
`katana --accounts 2 --disable-fee --invoke-max-steps 10000000`

contracts/src/components/labor_auction.cairo
```diff
@@ -0,0 +1,204 @@
...
+            )
+    }
+}
+
+
+#[derive(Component, Copy, Drop, Serde, SerdeLen)]
+struct LaborAuction {
+    #[key]
+    zone: u8,
+    decay_constant_mag: u128,
```
credence0x, line 68: Will decay constant and others be shared among zones?

contracts/src/systems/labor/purchase_labor.cairo
```diff
@@ -0,0 +1,287 @@
...
+        );
+
+        let zone = position.get_zone();
+        let mut labor_auction = get!(ctx.world, zone, (LaborAuction));
+        assert(labor_auction.per_time_unit != 0, 'Labor auction not found');
+
+        let mut labor_units_remaining = labor_units;
+        let mut total_costs: Felt252Dict<u128> = Default::default();
+
+        loop {
```
credence0x, line 48: is this loop needed?

contracts/src/systems/labor/build_labor.cairo
```diff
@@ -196,97 +196,75 @@ mod tests {
...
     }
 
 
-
     #[test]
     #[available_gas(300000000000)]
     fn test_build_labor_food() {
         let world = spawn_eternum();
 
+        let player_address = starknet::get_caller_address();
```
credence0x, line 255: Technically, the player_address is the test contract address. That's why we do `starknet::testing::set_contract_address` and not `starknet::testing::set_caller_address` before calling `world.execute`

contracts/src/systems/labor_auction/create_labor_auction.cairo
credence0x, line 1: I'm guessing calling this'll require admin permission. Else we should probably ensure it isn't called multiple times

contracts/src/components/labor_auction.cairo
```diff
@@ -0,0 +1,204 @@
...
+        self
+            .to_LinearVRGDA()
+            .get_vrgda_price(
+                // 1 period = 1 day
+                FixedTrait::new_unscaled(time_since_start / 86400, false), // time since start
+                FixedTrait::new_unscaled(self.sold, false), // amount sold
+            )
+    }
+
+    fn sell(ref self: LaborAuction, world: IWorldDispatcher) {
```
credence0x, line 101: The world dispatcher isn't doing anything

contracts/src/components/labor_auction.cairo
```diff
@@ -0,0 +1,204 @@
...
+
+
+#[derive(Component, Copy, Drop, Serde, SerdeLen)]
+struct LaborAuction {
+    #[key]
+    zone: u8,
+    decay_constant_mag: u128,
+    decay_constant_sign: bool,
+    per_time_unit: u128,
+    start_time: u64,
```
credence0x, line 71: All these are shared among all zones. Maybe they belong in a Config?

contracts/src/components/labor_auction.cairo
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    fn get_reverse_vrgda_price(self: @LinearVRGDA, time_since_start: Fixed, sold: Fixed) -> Fixed {
+        *self.target_price
+            * exp(
+                (*self.decay_constant * FixedTrait::new(1, true))
+                    * (time_since_start
+                        - self.get_target_sale_time(sold + FixedTrait::new(1, false)))
+            )
+    }
```
credence0x, line 60: This one is unsed. Intentional?

credence0x (commented):
awesome work!

contracts/src/components/labor_auction.cairo
aymericdelab: oups, totally right

r0man1337 (approved):

PR #150, opened 2023-09-28
Raid feature

credence0x:


credence0x:
@aymericdelab  i'd like you to please take a look

contracts/src/components/attack.cairo
```diff
@@ -0,0 +1,63 @@
+use eternum::alias::ID;
+use eternum::utils::coin_toss;
+
+#[derive(Component, Copy, Drop, Serde, SerdeLen)]
+struct Attack {
+    #[key]
+    entity_id: ID,
+    chance: u8,
```
aymericdelab, line 8: don't think we should call this chance, just value or something to make it generic even though we currently we do use it as proability in the attack system

contracts/src/components/attack.cairo
```diff
@@ -0,0 +1,63 @@
...
+    last_attack: u64
+}
+
+#[generate_trait]
+impl AttackImpl of AttackTrait{
+    
+    fn get_cost(self: @Attack) -> u8 {
```
aymericdelab, line 15: not sure what you mean by attack cost ? Attacks should be free as long as you are not in the cooloff period

contracts/src/components/movable.cairo
```diff
@@ -16,3 +19,61 @@ struct ArrivalTime {
     entity_id: u128,
     arrives_at: u64,
 }
+
+#[derive(Component, Copy, Drop, Serde, SerdeLen)]
+struct Travel {
```
aymericdelab, line 24: - feels like Movable and Travel serve the same purpose here. I don't see the point of having this info separated in 2 different components, if they will never be used one without the other.
- departure_time should not be stored because it can be computed from the arrival_time and coords
- storing departure_coord_x, departure_coord_y, last_stop_coord_x, last_stop_coord_y + Position component means you have 3 sets of coordinates to store. I think thiis could work with only 2 sets (as we discussed in discord)

contracts/src/components/movable.cairo
```diff
@@ -16,3 +19,61 @@ struct ArrivalTime {
...
+    // e.g if a caravan is traveling from A -> B -> A,
+    // last_stop_coord will be that of B.
+    last_stop_coord_x: u32,
+    last_stop_coord_y: u32
+}
+
+
+#[generate_trait]
+impl TravelImpl of TravelTrait {
+    fn live_location(self: @Travel, destination: Coord, timestamp: u64) -> Coord {
```
aymericdelab, line 41: that's a great idea to implement this trait, think this should be implemented for movable though

contracts/src/components/position.cairo
```diff
@@ -11,6 +11,75 @@ struct Coord {
     y: u32
 }
 
+#[generate_trait]
+impl CoordImpl of CoordTrait {
+
+    // todo@credence revisit this
+    fn interpolate(self: Coord, other: Coord, percent: u32) -> Coord {
```
aymericdelab, line 18: can you explain in a few words the goal of this function ?

contracts/src/components/attack.cairo
credence0x: get_cost there is to calculate the shekel amount that will be paid for attack when creating a mercenary. E.g if you want a 10% chance of victory during attacks, you pay 1 * 40 shekels. If you want 50 percent chance, you pay 5 * 40 shekels.

Calling the raid system itself doesn't require additional cost

contracts/src/systems/raid.cairo
```diff
@@ -0,0 +1,128 @@
...
+            loop {
+                match resource_types.pop_front() {
+                    Option::Some(resource_type) => {
+                        let adversary_resource = get!(ctx.world, adversary_entity_id, Resource);
+
+                        resource_total_weight += WeightConfigImpl::get_weight(
+                                ctx.world, *resource_type, adversary_resource.balance
+                            );
+
+                        if resource_total_weight < raiding_entity_total_capacity {
```
aymericdelab, line 83: should be able to take the max amount carraible even if total amount is bigger weight that he can carry

contracts/src/systems/raid.cairo
```diff
@@ -0,0 +1,128 @@
...
+              
+        if raiding_entity_attack.launch() { 
+            // successful attack
+
+            let raiding_entity_quantity = get!(ctx.world, raiding_entity_id, Quantity);
+            let raiding_entity_total_capacity = raiding_entity_capacity.weight_gram * raiding_entity_quantity.get_value();
+
+            let mut resource_total_weight = 0;
+
+            loop {
```
aymericdelab, line 74: in my mind, original goal was to be able to steal resourecs, only if the caravan/donkeys are all dead

contracts/src/components/position.cairo
credence0x: Something is traveling from A -> B. Given  A, B and the percentage of the journey completed, it calculates the exact coordinate that the travelling entity is at. i.e( somewhere between A and B )

I might rewrite it entirely so no need to look too deeply but that's what it does

contracts/src/systems/raid.cairo
```diff
@@ -0,0 +1,128 @@
...
+
+                    Option::None => {
+                        break;
+                    }
+                };    
+            };
+
+            
+            // todo@credence move hardcoded damage to config
+            // random damage from 1 - 50
```
aymericdelab, line 108: not sure if we need randomness here, maybe just removing the attack value from the health should suffice for now. We could check with fomentador to make something more fun

contracts/src/systems/raid.cairo
```diff
@@ -0,0 +1,128 @@
+#[system]
+mod Raid {
```
aymericdelab, line 2: thinking of calling this AttackTravelingEntity rather than raiding

contracts/src/components/movable.cairo
credence0x: If we recompute departure_time, it might be incorrect. e.g if the traveling entity uses a road. Unless we want to store that  a road was used

contracts/src/systems/mercenary.cairo
```diff
@@ -0,0 +1,121 @@
...
+    use eternum::components::quantity::{Quantity, QuantityTrait};
+
+    use eternum::constants::ResourceTypes;
+    
+    use dojo::world::Context;
+
+    use core::traits::Into;
+
+
+    fn execute(ctx: Context, entity_id: u128, attack_chance: u8, mut transport_unit_ids: Array<u128>) {
```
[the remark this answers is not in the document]
credence0x: 'attack_chance as input' - mistake. Forgot to delete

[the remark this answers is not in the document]
credence0x: > You should create a mercenary on your realm with a certain capacity and attack that are constant. No link with other transport units.

Oh. What about movement. They can't travel if there is no movable comp. Will that be fixed too?

contracts/src/systems/raid.cairo
credence0x: I'm not sure i understand. What do you mean by max amount carriable? Also they can carry more than their capacity? Is there any need for capacity then?

contracts/src/systems/mercenary.cairo
```diff
@@ -0,0 +1,121 @@
...
+        set!(ctx.world, (
+                attack, 
+                Owner {
+                    entity_id: mercenary_id, 
+                    address: ctx.origin
+                },
+                Movable {
+                    entity_id: mercenary_id, 
+                    sec_per_km: average_speed, 
+                    blocked: false,
```
credence0x, line 107: https://github.com/BibliothecaDAO/eternum/pull/150#discussion_r1341314819

contracts/src/systems/raid.cairo
credence0x: We could but i think that'll be limiting. With what we have here, we could basically just add the `Health` component to anything and it becomes raidable. Whether travelling or not. It doesn't actually have to be movable. If it isn't movable, its live_location would just be its `Position`

PR #183, opened 2023-10-17
update trade systems

credence0x:


aymericdelab:
Also, we need to implement the roundtrip feature with new Movable component:
```
struct Movable {
      sec_per_km,
      blocked: boolean,
      roundtrip: boolean,
      intermediate_position_x,
      intermediate_position_y,
}
```

PR #192, opened 2023-10-19
feat: notifications claim button

r0man1337:
added claim button to notifications about arrived caravans


aymericdelab:
build issue

PR #204, opened 2023-10-23
add C&C in eternum

FrostStarBook:
add C&C contract and Front-end integrated


sdk/packages/eternum/src/types/index.ts
```diff
@@ -3,6 +3,13 @@ import { Account, num } from "starknet";
 interface SystemSigner {
   signer: Account;
 }
+export interface MintCC extends SystemSigner {
```
ponderingdemocritus, line 6: this doesn't need to exist - just use SystemSigner

sdk/packages/eternum/src/constants/contracts.ts
```diff
@@ -11,6 +11,7 @@ export const DEV_CONTRACTS = {
   TRAVEL_SYSTEMS: "0x70717be365c143d9f4ae207e420d0c3525a7c79197a20e6e63e4dec0b1b26cd",
   TEST_REALM_SYSTEMS: "0xee850f8d8ded18763d23e35d3592f9549081124bee0c32ffb2ef355deb1b68",
   TEST_RESOURCE_SYSTEMS: "0x17505387265fa24a4846cb3b3207aa92c3e6f918f74883a43be959265d19e44",
+  CC_CONTRACT_ADDRESS:"0x1e26e39c8d6d9d03179426d6a95190001b72a75645da0a3cff264c936bd3f1",
```
ponderingdemocritus, line 14: why are production and development the same?

contracts/src/systems/cc_map/interface.cairo
```diff
@@ -0,0 +1,41 @@
+// ---------------------------------------- interface -------------------------------------
+#[starknet::interface]
+trait CCInterface<TContractState> {
```
ponderingdemocritus, line 3: this should be exposed in the CC repo and imported, it shouldn't exist within eternum repo

contracts/src/systems/cc_map/contract.cairo
```diff
@@ -0,0 +1,141 @@
+// ---------------------------------------- contract --------------------------------------
```
ponderingdemocritus, line 1: this shouldn't exist in this repo either

sdk/packages/eternum/src/types/index.ts
FrostStarBook: np

contracts/src/systems/cc_map/contract.cairo
FrostStarBook: This is dojo::contract. We are not able to import the package by dependency, we need to have an implementation in the repo.

sdk/packages/eternum/src/constants/contracts.ts
FrostStarBook: How to get Production Contract Address

ponderingdemocritus:
I think this should exist on an alternatively hosted client - so we decentralise the clients

ponderingdemocritus:
I have added instructions in the client readme - can you spin an alt client up?

FrostStarBook:
> I have added instructions in the client readme - can you spin an alt client up?

I've seen it, it's written very clearly. We will remake one. The front end just needs a redirect button or something like that. The contract part will submit a PR.

ponderingdemocritus:
It makes sense to maintain a fork of this rather than PR. This way the C&C client can move faster than waiting for a merge.

FrostStarBook:
> It makes sense to maintain a fork of this rather than PR. This way the C&C client can move faster than waiting for a merge.

i see

PR #211, opened 2023-10-31
Combat

credence0x:
This allows for basic combat allowing for soldiers to be created and grouped as well as allowing for attack and stealing realms resources. The main components  used are Attack, Defence and Health. Soldiers can be viewed as individual raiding units until they are grouped. it's just that when they are grouped they stand a higher chance of survival.

They can be grouped to perform 2 duties.
1. Raid: if they have this duty, they are still free to travel, raid and return
2. TownWatch: when they have this duty, we ensure the soldier is at the realm and block them from moving. Then we add their strength to the town watch.

Soldiers can also be ungrouped when they are anywhere. So if you get to a realm and decide you need to split up your raiding unit or army, you can ungroup then create maybe 2 different groups. You may even decide to leave them ungrouped and control soldiers individually. The only caveat is that during group creation, soldiers must not have any time in their inventory. Also during ungrouping, the group should also not have any item in inventory. They must have offloaded it somewhere


contracts/src/systems/combat/contracts.cairo
```diff
@@ -0,0 +1,398 @@
...
+
+    #[external(v0)]
+    impl SoldierSystemsImpl of ISoldierSystems<ContractState> {
+
+        fn create_soldiers( 
+            self: @ContractState, world: IWorldDispatcher, 
+            entity_id: u128, quantity: u128
+        ) {
+
+            // q: check if entity is a realm?
```
ponderingdemocritus, line 32: yes

contracts/src/systems/combat/contracts.cairo
```diff
@@ -0,0 +1,398 @@
...
+                    },
+                    Attack {
+                        entity_id: soldier_id,
+                        value: 100
+                    },
+                    Defence {
+                        entity_id: soldier_id,
+                        value: 100
+                    },
+                    Quantity {
```
ponderingdemocritus, line 70: does this need to exist?

credence0x: Hmm I think so. Quantity value is the only way we can differentiate between single soldiers and groups of them (like Raiders)

ponderingdemocritus: ok right - i understand now

contracts/src/systems/combat/contracts.cairo
```diff
@@ -0,0 +1,398 @@
...
+    use eternum::constants::{
+        WORLD_CONFIG_ID, SOLDIER_ENTITY_TYPE
+    };
+
+
+    #[external(v0)]
+    impl SoldierSystemsImpl of ISoldierSystems<ContractState> {
+
+        fn create_soldiers( 
+            self: @ContractState, world: IWorldDispatcher,
```
ponderingdemocritus, line 28: we should pass in realm_id maybe - and use its location to build

credence0x: the entity_id would be the realm id

credence0x: Since it's only meant to be called by realms, I'll just change the name to realm_id

aymericdelab: we should try to avoid using realm_id and use realm_entity_id instead in the contracts. Since realm_id already exists in Realm.

There might still be a couple place in contract where we use realm_id but we should refactor

aymericdelab (approved):

contracts/src/models/combat.cairo
```diff
@@ -0,0 +1,42 @@
...
+#[derive(Model, Copy, Drop, Serde)]
+struct Attack {
+    #[key]
+    entity_id: u128,
+    value: u128
+}
+
+
+#[derive(Model, Copy, Drop, Serde)]
+struct Defence {
```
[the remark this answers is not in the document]
credence0x: Given the amount of effort it required for Britain to colonise Nigeria, I do not think the Queen Elizabeth (God bless her soul) would have been happy to see me spell it as `Defense`. It would seem like all that effort had been in vain. So I hope you see the reason for my decision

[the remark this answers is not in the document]
aymericdelab: god bless the queen 🇬🇧 👑

PR #213, opened 2023-11-01
feat: ui features

r0man1337:
1. UI update for VRGDA rates
image

2. Reactive farms count based on labor
image

3. Direct offer UI update
image

4. Labor model tooltips update
image


ponderingdemocritus:
one conflict then lets merge

PR #237, opened 2023-11-15
make transfer systems compatible with inventory

credence0x:
- resources transferred from realm to any other entity goes to the entity's inventory
- the entities(e.g caravans) now offload resources by calling the `transfer_item` function instead of `offload_chest`. they work very similarly
- minor update to `trade.create_order` API


ponderingdemocritus:
have change to 0.3.0 @credence0x can you check merge conflicts, then lets merge

aymericdelab:
@credence0x  i think we can close this since it's in feat-banks

PR #239, opened 2023-11-16
fix: remove food and shekels from pillage realm

aymericdelab:


aymericdelab:
@ponderingdemocritus  lgtm

PR #253, opened 2023-11-20
feat: add address name

aymericdelab:
Added possibility to add a name to your address.

Prompted to enter address if you haven't when sign in.

If you already have a name, you cannot change it.


ponderingdemocritus:
is this breaking change? should be push this to 0.3.0?

aymericdelab:
no breaking change, would work with current version, would just need to sozo --release migrate when merged into main

ponderingdemocritus:
ok lets ship it!

ponderingdemocritus (approved):

PR #327, opened 2023-12-21
Guild Components

zhaowei666:
1. Create guild flow
2. update chat config
3. add chat notifications


aymericdelab:
gm lords, awesome idea, having a separate chat for smaller groups is a great addition. At this state of the game priority should be to focus on chat between Orders. 
Do you think you could adapt this to be able to talk between realms of the same other rather than of the same guild ?

PR #339, opened 2024-01-10
feat: caravan disassembly and free transport unit return

credence0x:
- added a way to disassemble caravan into its transport units in `caravan_systems.disassemble`
- added a way to return unused free transport units in `transport_unit_systems.return_free_units`


aymericdelab (approved):
awesome work

PR #382, opened 2024-03-10
Bug fixes for hex travel, map exploration and army to army combat

credence0x:
- allows players explore more than one hex and hold multiple items in inventory before travelling back to realm
- fix army combat bugs
- fix hex travel bugs


credence0x:
PR for #383

ponderingdemocritus:
could you update CI. I believe dojo version issue.

credence0x:
> I believe dojo version issue.

made a PR. CI still won't work though because of https://github.com/dojoengine/dojo/issues/1646

contracts/src/systems/combat/contracts.cairo
aymericdelab, line 1: im thinking maybe long term, wouldn't it be better to add defence, attack to a realm depending on the number of armies that you attach to it ? rather than having this townwatch system ? just a thought

aymericdelab (changes requested):

contracts/src/systems/combat/contracts.cairo
aymericdelab: not a change for this pr though

contracts/scripts/set_config.sh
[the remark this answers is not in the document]
credence0x: though i added it here but accidentally added it to wrong commit. so i've merged the commit here and you can check again

contracts/src/systems/combat/contracts.cairo
credence0x: yeah i think it would make sense to just attach attack and defence to realm. but i remember that there were good reasons i decided against it. will look into it again if the need arises

contracts/scripts/set_config.sh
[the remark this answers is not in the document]
aymericdelab: awesome

contracts/src/systems/combat/contracts.cairo
aymericdelab: yeah lets come back to this later

aymericdelab (approved):

PR #386, opened 2024-03-11
CI: update dojo version in CI

credence0x:
- update CI dojo version
- test won't run in CI though because of this issue https://github.com/dojoengine/dojo/issues/1646 but to confirm that it works, `sozo build` should work in CI and `sozo test` should be attempted in CI


credence0x:
PR for #387

aymericdelab (approved):

Issue #375, opened 2024-03-08
Allow for raiders to explore more than once before offloading resources


aymericdelab:
**Is your feature request related to a problem? Please describe.**
Currently the raiders can only explore once, and need to offload their discovered resources in their realm, before being able to explore again.

**Describe the solution you'd like**
In the contracts, they should be able to carry more resources. Each resource being on a new spot in their inventory.
In the client, you need to change this line:
https://github.com/BibliothecaDAO/eternum/blob/4c097a4c944a128b267e5fbb01f613d0dd0e3fe8/client/src/components/worldmap/ChooseActionPopup.tsx#L96

credence0x:
resolved by #392

PR #392, opened 2024-03-12
client: allow for multiple hex exploration before return to realm

credence0x:
- Allow for multiple hex exploration before return to realm
- batch calls to transfer inventory items to realm to avoid hitting cairo step limit
- only enable explore button when army has enough capacity to carry reward

### DIFF
--- a/client/src/components/worldmap/ChooseActionPopup.tsx
+++ b/client/src/components/worldmap/ChooseActionPopup.tsx
@@ -93,7 +93,7 @@ export const ChooseActionPopup = ({}: ChooseActionPopupProps) => {
               <Button
                 variant="primary"
                 size="md"
-                disabled={hasResources || isTraveling}
+                disabled={isTraveling}
                 onClick={onExplore}
                 className=""
               >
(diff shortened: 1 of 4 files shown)


credence0x:
resolves #375

aymericdelab (changes requested):
here we'd need to check if it has enough capacity to carry the extra resources that it could get from explore. So calculate current weight of the inventory and total capacity of the army

credence0x:
> here we'd need to check if it has enough capacity to carry the extra resources that it could get from explore. So calculate current weight of the inventory and total capacity of the army.
> 
> There's a function called `getTotalResourceWeight` and raiders capacity should already be in raider object (CombatInfo interface)



> here we'd need to check if it has enough capacity to carry the extra resources that it could get from explore. So calculate current weight of the inventory and total capacity of the army.
> 
> There's a function called `getTotalResourceWeight` and raiders capacity should already be in raider object (CombatInfo interface)

updated. decided against using `getEntitiesCombatInfo` because it used too many queries and would contain too many irrelevant info.

aymericdelab (approved):

PR #394, opened 2024-03-13
feat: fix road bugs and allow create any road

### LINKED ISSUE #275: Arbitrary way to build roads
We need a way to build arbitrary roads between Realms. Right now it's only possible when clicking on an order.

### LINKED ISSUE #353: Same roads are repeating twice in roads tab
Same roads are repeated twice in the roads card of the realm. Maybe because I created the same route, separately from each of the realms to the other. 

image

### LINKED ISSUE #354: Resource shortage not shown on the road replenish screen
When I'm adding usage to a road, I see as if I have all the resources to do it. But after checking each of them, I'm actually out of wood but that shortage is not shown on the road usage adding screen. It should probably stop me from being able to click on add usage and show the wood in red (short). 

image

aymericdelab:
resolves #354 
resolves #353
resolves #275


credence0x (approved):
lgtm

PR #395, opened 2024-03-13
client: feat: show enemy army on map

### LINKED ISSUE #376: See enemy armies in the client
**Is your feature request related to a problem? Please describe.**
Currently we only see our own armies on the map. 

**Describe the solution you'd like**
Let's add a new EnemyArmies component that fetches all the enemy armies and displays them on the map. They should have a different color from your own armies. And should not trigger anything when clicked on it.

credence0x:
- Show enemy army on the world map
resolves #376


credence0x:
resolves #376

aymericdelab (approved):

PR #397, opened 2024-03-14
feat: show travel animation for ennemies and exploring

### LINKED ISSUE #396: add animations for all travel (exploration and travel hex)
Currently, only own raider armies have a "travel" animation when they call travel_hex, we should show the animation for any army travelling but also any army exploring (since they travel as well). 

Best way would be to use the events from the travel function, to show animation every time they are emitted for any army.

aymericdelab:
resolves #396


aymericdelab:
@credence0x fixed the issue with the disappearing armies

credence0x (approved):
lgtm

credence0x (approved):
lgtm

PR #421, opened 2024-03-19
fix biomes

aymericdelab:
fixed seasonal forest but seems like same issue would arise in other biomes as well.


r0man1337:
Fixed all the other biomes

r0man1337 (approved):

PR #488, opened 2024-03-28
Optimize Cairo Tests - Part 1

gianalarcon:
Partially closes 436 

## Description 
To enhance the efficiency of current Cairo tests, I suggest, first to fully upgrade the contracts to Cairo  `2.5.0` and dojo `v0.6.0-alpha.4`. After that, we can do some major refactoring/enhancement to reduce the test execution time

In this PR, I propose updating the syntax of certain attributes and incorporating macros for improved clarity and functionality:

- [x] Transition to using  `assert!`, `assert_eq!`, `assert_ne!` macros 
- [x] Enhance the utilization of `assert!` on boolean assertions
- [x] Remove the optional `available_gas` attribute
- [x] Update the `should_panic` attribute 

It's important to note that this PR maintains the existing program and test logic intact. Subsequent PRs will focus on refining the test setup through refactoring efforts.


gianalarcon:
> thanks for the PR ! are we using a different cairo formater as you are ? hard to see what changes you made because of all the formatting

Oh yes. I will fix the formatter to make it clear. For now I just replaced some `assert` built in fn with `assert!`macro. I will still upgrading some sintax to cairo 2.54 and dojo v0.6.4

aymericdelab:
what's the reason behind removing #[available_gas(30000000)] ? 
is this not needed anymore ? 
seems like it's still in use here: https://book.dojoengine.org/tutorial/onchain-chess/3-test

gianalarcon:
> what's the reason behind removing #[available_gas(30000000)] ? 
> is this not needed anymore ? 
> seems like it's still in use here: https://book.dojoengine.org/tutorial/onchain-chess/3-test

In previous versions it was mandatory to use #[available_gas(30000000)], but now it is not, we can skip it. 

Oh yes. We have to update the tests content of dojo book as well. Noted

gianalarcon:
@aymericdelab @credence0x @ponderingdemocritus. I am done with this PR, basically it is a syntax update, please check. 
After this one, I plan to create a `PR - Part 2` where I will do more refactoring/enhancement.

ponderingdemocritus:
Could you fix the CI @gianalarcon

gianalarcon (commented):
> Could you fix the CI @gianalarcon  

My bad, It is fixed now.

aymericdelab:
- tests seems to still take around 35 minutes to run so don't think the performance issue has been solved yet.
- also, only 96/168 tests are passing now, while 166/168 were passing before that

gianalarcon:
> - tests seems to still take around 35 minutes to run so don't think the performance issue has been solved yet.
> - also, only 96/168 tests are passing now, while 166/168 were passing before that 

Yes. What a pity. I will try another way on a different PR

ponderingdemocritus:
@gianalarcon the only way we can solve this giant test suite is with starknet foundry - which currently isn't supported by dojo. It is in the works. Going to close this for now, as a bunch has changed, so if you do tackle in future just start again.

PR #627, opened 2024-04-27
Mute whoosh sound when panning

Await-0x:
## **Type**
enhancement

___

## **Description**
- Introduced an optional `muteSound` property to the camera controls, allowing the control of sound effects during camera panning.
- The whoosh sound effect during camera transitions can now be muted by setting the `muteSound` property to true.
- This enhancement improves user experience by providing control over audio feedback during navigation.

___

## **Changes walkthrough**
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>Camera.tsx</strong><dd><code>Add Mute Sound Option to Camera Controls</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

client/src/ui/utils/Camera.tsx
<li>Added <code>muteSound</code> optional property to <code>position</code> and <code>target</code> in <code>Props</code> <br>interface.<br> <li> Conditionally play the fly sound based on the <code>muteSound</code> property.<br> <li> Updated state setting for camera position and target to include <br><code>muteSound</code>.<br>

</details>
    

  </td>
  <td><a href="https://github.com/BibliothecaDAO/eternum/pull/627/files#diff-4dfb619ce95fdd0f7db51d19e40b93e26c80636b93e0d277a21c7353c083aa6d">+7/-3</a>&nbsp; &nbsp; &nbsp; </td>
</tr>                    
</table></td></tr></tr></tbody></table>

___

> ✨ **PR-Agent usage**:
>Comment `/help` on the PR to get a list of all available PR-Agent tools and their descriptions

### DIFF
--- a/client/src/ui/utils/Camera.tsx
+++ b/client/src/ui/utils/Camera.tsx
@@ -15,12 +15,14 @@ interface Props {
     y: number;
     z: number;
     transitionDuration?: number;
+    muteSound?: boolean;
   };
   target: {
     x: number;
     y: number;
     z: number;
     transitionDuration?: number;
+    muteSound?: boolean;
   };
 }
 
@@ -95,7 +97,9 @@ const CameraControls = ({ position, target }: Props) => {
 
   useEffect(() => {
     cameraAnimate();
-    playFly();
+    if (!position.muteSound && !target.muteSound) {
+      playFly();
+    }
   }, [target, position]);
 
   const clampPan = useCallback(
@@ -105,8 +109,8 @@ const CameraControls = ({ position, target }: Props) => {
       _v.sub(e?.target.target);
       camera.position.sub(_v);
 
-      // setCameraPosition(camera.position.clone());
-      // setCameraTarget(ref.current.target.clone());
+      setCameraPosition({ ...camera.position.clone(), muteSound: true });
+      setCameraTarget({ ...ref.current.target.clone(), muteSound: true });
     },
     [camera],
   );


client/src/ui/utils/Camera.tsx
```diff
@@ -105,8 +109,8 @@ const CameraControls = ({ position, target }: Props) => {
       _v.sub(e?.target.target);
       camera.position.sub(_v);
 
-      // setCameraPosition(camera.position.clone());
-      // setCameraTarget(ref.current.target.clone());
+      setCameraPosition({ ...camera.position.clone(), muteSound: true });
+      setCameraTarget({ ...ref.current.target.clone(), muteSound: true });
```
r0man1337, line 113: I'm not sure if these two lines should be here. Let's remove them.

**cameraPosition** and **cameraTarget** in the zustand store are used to trigger automatic transition from current camera position to the new one E.g. wen user clicks "show on map" - he will see an animated transition to that entity. So **cameraPosition** and **cameraTarget** do not always storing current actual camera values, but they are used only once at the moment of transition (and usually they are not changed directly, but by calling a function like moveCameraToRealmId)

And clampPan function is just restricts user to move outside of the map bounds, there's no need to set camera position and target every time the user pans. Also I see one improvement that we can make - let's add a condition that the clampPan is executed only if the user is really trying to go off the map bounds, instead of clamping every time user pans, even if the camera does not go off the bounds.

Await-0x: You guys know best but

if we want to have a compass then we need to store the current camera position and target somewhere. The clampPan function triggers when the user rotate/move the camera. Problem was "flySound" got triggered with change to cameraPosition.

If you remove those lines I would suggest also removing the compass

r0man1337: > You guys know best but
> 
> if we want to have a compass then we need to store the current camera position and target somewhere. The clampPan function triggers when the user rotate/move the camera. Problem was "flySound" got triggered with change to cameraPosition.
> 
> If you remove those lines I would suggest also removing the compass

Got your point. I moved code for calculating direction of compass to MapControls onChange handler and added throttling for it. Now it should work fine. There were two reasons why I didn't want to set camera pos & target - 1) this a trigger for statring camera animation 2) onChange is called very frequently and could lead to performance drops if combined with reassigning variables and recalculating direction value

Await-0x: Nice, looks good!

PR #798, opened 2024-05-22
Fixes compass element

### LINKED ISSUE #705: [client] improve Compass performance
Currently Compass component is commented out as the fps are dropping when it's there. We need a more performant way to show it in the client.

cwastche:
- Re-renders the element only on camera movement, saving on GPU mem usage 
- Directions now correctly applied to the entire 360° range
Closes #705


aymericdelab:
since map direction is now fixed, do we still need this ? @ponderingdemocritus

ponderingdemocritus:
agreed - lets leave this

PR #809, opened 2024-05-24
feat: add hyperstructures UI

edisontim:
This PR adds: 
- Possibility to create hyperstructure from the world map 
- Possibility to see existing hyperstructures
- Possibility to contribute to hyperstructures


client/src/hooks/store/_mapStore.tsx
ponderingdemocritus, line 1: do we need to set this is state? The structures exist in recs already.

client/src/hooks/store/useLeaderBoardStore.tsx
```diff
@@ -1,123 +1,234 @@
...
+  [ResourcesIds.DeepCrystal]: 20.98,
+  [ResourcesIds.Ignium]: 29.15,
+  [ResourcesIds.EtherealSilica]: 30.95,
+  [ResourcesIds.TrueIce]: 36.06,
+  [ResourcesIds.TwilightQuartz]: 45.18,
+  [ResourcesIds.AlchemicalSilver]: 53.92,
+  [ResourcesIds.Adamantine]: 91.20,
+  [ResourcesIds.Mithral]: 135.53,
+  [ResourcesIds.Dragonhide]: 217.92,
+  [ResourcesIds.Earthenshard]: 20.98,
```
ponderingdemocritus, line 47: this should be more, as it can only be found on the map. Change to 400.

client/src/hooks/store/useLeaderBoardStore.tsx
```diff
@@ -1,123 +1,234 @@
...
 
 interface LeaderboardStore {
   loading: boolean;
   setLoading: (loading: boolean) => void;
-  playerResourceLeaderboard: PlayerResourceLeaderboardInterface[];
-  orderResourceLeaderboard: OrderResourceLeaderboardInterface[];
-  setResourceLeaderboards: (
-    playerResourceLeaderboard: PlayerResourceLeaderboardInterface[],
-    orderResourceLeaderboard: OrderResourceLeaderboardInterface[],
+  playerPointsLeaderboard: PlayerPointsLeaderboardInterface[];
```
ponderingdemocritus, line 84: do we need to have this in the state? Could we just fetch with recs?

client/src/ui/components/models/buildings/worldmap/Structures.tsx
```diff
@@ -0,0 +1,88 @@
...
+      return {
+        col: position!.x,
+        row: position!.y,
+        type: type as StructureType,
+        entity: entity,
+      };
+    });
+  }, [builtStructures.length]);
+
+  useEffect(() => {
```
ponderingdemocritus, line 45: this could be a useMemo probably with recs query inside. Might be better to move this into hook aswel.

client/src/hooks/store/useLeaderBoardStore.tsx
aymericdelab: this is the same amount as in the contract currently (same as ruby). The weight will be 1/x of this though since this represents the rarity

aymericdelab: later on, this should be in config and it should be set in contract config as well rather than being constants in constants.cairo

client/src/hooks/store/useLeaderBoardStore.tsx
aymericdelab: could be computation heavy to calculate leaderboard for all players so might be better to set it at the beginning and then have a refresh button, i think zustand store is good in this case

client/src/ui/components/models/buildings/worldmap/Structures.tsx
aymericdelab: agreed

client/src/hooks/store/_mapStore.tsx
aymericdelab: if you're talking about the existing structures, it's practical to have it in zustand state because we're using both for 3d models, but also when trying to place a hyperstructure, every time it hovers on a hex we want to know if there's already a structure to show "red" "green" model color

client/src/hooks/store/useLeaderBoardStore.tsx
aymericdelab: currently you can find them on the map + players who find a shards mine have infinite production of it, so i'm not sure it's necessarily more rare. Depends the prob of finding a shards mine

ponderingdemocritus:
- Hyperstructure part construction doesn't show after a contribution has been made
- Mines not showing on world map
- Need to flatten the hexception view once a mine/hyperstructure has been built

aymericdelab (approved):

PR #830, opened 2024-05-28
feat: compute all config rather than store in state

EjembiEmmanuel:


ponderingdemocritus:
closing as doesn't seem active

PR #949, opened 2024-06-17
Battle fixes

credence0x:
- force remove enemy from current battle if they lost. This prevents any error that doesnt allow a structure to be pillaged because their resources are locked in battle. even though theyve lost
- remove "send home after pillage" feature


edisontim (approved):
lgtm

PR #1001, opened 2024-06-24
feat: refactor battle view

edisontim:


client/src/ui/modules/military/battle-view/Battle.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { ComponentValue } from "@dojoengine/recs";
+import { motion } from "framer-motion";
+import { useState } from "react";
+import { BattleActions } from "./BattleActions";
+import { BattleProgressBar } from "./BattleProgressBar";
+import { BattleSideView } from "./BattleSideView";
+import { LockedResources } from "./LockedResources";
+import { TopScreenView } from "./TopScreenView";
+
+export interface Health {
```
aymericdelab, line 16: should we move this in eternum sdk ?

edisontim: Health will be removed from the `BattleView` so it can be removed

aymericdelab (approved):

PR #1038, opened 2024-06-28
fix: troops in battle

edisontim:
Fix bug where battle manager was updating the object of the recs by subtracting health and troops


edisontim:
Wrong branch

PR #1285, opened 2024-08-16
contracts: limit battle reward given by winner's capacity

### LINKED ISSUE #1247: [contracts] Entity capacity not enough
I'm getting this error while trying to leave a battle that I won 
Screenshot 2024-08-14 at 12 35 04

Screenshot 2024-08-14 at 12 35 15

Screenshot 2024-08-14 at 12 34 24

credence0x:
resolves #1247


contracts/src/systems/resources/contracts.cairo
```diff
@@ -269,6 +269,48 @@ mod resource_systems {
...
+            // only add to balance if receiver can carry weight
+            let (resource_type, resource_amount) = resource;
+            let mut total_resources_weight = 0;
+            total_resources_weight += WeightConfigCustomImpl::get_weight(world, resource_type, resource_amount);
+            let mut recipient_weight: Weight = get!(world, recipient_id, Weight);
+            let recipient_capacity: Capacity = get!(world, recipient_id, Capacity);
+            let recipient_quantity: Quantity = get!(world, recipient_id, Quantity);
+
+            recipient_weight.value += total_resources_weight;
+            if !recipient_capacity.is_capped() || recipient_capacity.can_carry(recipient_quantity, recipient_weight) {
```
aymericdelab, line 289: recipient quantity is currently with resource precision, while weight is in grams per true unit (not precision unit) so the total capacity in can_carry will be too high by a factor of x1000

contracts/src/systems/resources/contracts.cairo
```diff
@@ -269,6 +269,48 @@ mod resource_systems {
...
+
+            // only add to balance if receiver can carry weight
+            let (resource_type, resource_amount) = resource;
+            let mut total_resources_weight = 0;
+            total_resources_weight += WeightConfigCustomImpl::get_weight(world, resource_type, resource_amount);
+            let mut recipient_weight: Weight = get!(world, recipient_id, Weight);
+            let recipient_capacity: Capacity = get!(world, recipient_id, Capacity);
+            let recipient_quantity: Quantity = get!(world, recipient_id, Quantity);
+
+            recipient_weight.value += total_resources_weight;
```
aymericdelab, line 288: why do you increase the recipient weight here ?

contracts/src/systems/resources/contracts.cairo
```diff
@@ -269,6 +269,48 @@ mod resource_systems {
...
+            if !recipient_capacity.is_capped() || recipient_capacity.can_carry(recipient_quantity, recipient_weight) {
+                recipient_weight.value -= total_resources_weight;
+                recipient_weight.add(recipient_capacity, recipient_quantity, total_resources_weight);
```
aymericdelab, line 291: when you add here pay attention that recipient quantity is in 1000x precision while the capacity and weight is in gram per true unit (not x1000) so you need to divide quantity by 1000 before checking assert_can_carry in the weight add

contracts/src/systems/resources/contracts.cairo
credence0x: so that the recipient_weight in the next line `recipient_capacity.can_carry(recipient_quantity, recipient_weight)` includes the weight of the new resources. we're essentially asking, can recipient carry old weight + new weight

contracts/src/systems/resources/contracts.cairo
credence0x: does this solve that
https://github.com/BibliothecaDAO/eternum/blob/main/sdk/packages/eternum/src/config/index.ts#L221-L226

aymericdelab: we're gonna change that and add factor of x1000 because in config it's in kg but we want grams

contracts/src/systems/resources/contracts.cairo
aymericdelab: ok makes sense

edisontim:
@credence0x can you have a look at @aymericdelab's comments with the new values of `setCapacityConfig`? We've changed the numbers there so that they're in grams and set them to reflect the true units (not precision)

aymericdelab (approved):

Issue #1297, opened 2024-08-16
[threejs] remove explored hex if tx reverts


aymericdelab:

aymericdelab:
@r0man1337 i've updated the system  in #1299 so that we the onUpdate function knows if the update is an explored hex or an unexplored hex (if tx reverts and optimistic is removed so an explored hex becomes unexplored). 

Could you check how to actually remove the explored hex from the worldmap ?

r0man1337:
Fixed.

PR #1299, opened 2024-08-16
add explore boolean in system update

aymericdelab:


aymericdelab:
fixes #1297

aymericdelab:
would still need a way to delete explored hex in the model

edisontim (approved):

PR #1302, opened 2024-08-17
Pillage reward and army stamina change

### LINKED ISSUE #1276: [contract] Pillaging should use up the army's stamina
Currently, you can pillage an infinite amount of times (or until either army is dead).

1. We should change that to the raiding army using up ALL of its remaining stamina.
2. The amount of stamina you have when raiding should affect the amount of resources you get when successful

Ex: 
if at 100% stam, get 100% of resources
if at 50% stam, get 50% of resources

credence0x:
resolves #1276


contracts/src/models/stamina.cairo
```diff
@@ -62,9 +61,15 @@ impl StaminaCustomImpl of StaminaCustomTrait {
...
         assert(maxes.len() > 0, 'No troops in army');
 
-        self.amount = maxes.min().unwrap();
-        self.last_refill_tick = armies_tick_config.current();
+        maxes.min().unwrap()
+    }
+
+    fn refill(ref self: Stamina, world: IWorldDispatcher) {
+        self.amount = self.max(world);
         self.sset(world);
```
edisontim, line 69: what's this function? Can we give a more explicit name to it please?

edisontim (approved):
lgtm

contracts/src/systems/combat/contracts.cairo
```diff
@@ -958,8 +965,12 @@ mod combat_systems {
                 };
             }
 
-            let mut destroyed_building_category = BuildingCategory::None;
+            // drain stamina
+            army_stamina.amount = 0;
```
aymericdelab, line 969: if you set the amount to 0 but don't change the last_refill_tick doesn't it mean that it could still be full ? maybe use the refill function first before doing any stamina changes ? also use sset method to keep consistency

contracts/src/models/stamina.cairo
credence0x: lol you wrote it. refill stamina. maybe reset?

contracts/src/systems/combat/contracts.cairo
credence0x: updated

aymericdelab: thank u ser

PR #1313, opened 2024-08-18
feat: add generic sozo events

edisontim:


contracts/src/systems/combat/contracts.cairo
```diff
@@ -686,6 +687,27 @@ mod combat_systems {
...
+                    id,
+                    event_id: EventType::BattleStart,
+                    timestamp: starknet::get_block_timestamp(),
+                    data: EventData::BattleStart(
+                        BattleStartData {
+                            attacker,
+                            defender,
+                            attacker_army_entity_id: attacking_army_id,
+                            defender_army_entity_id: defending_army_id,
+                            duration_left: 100,
```
cwastche, line 706: Do we even need this attribute in battleStartData ?

ponderingdemocritus: We could compute it in the client or here. It might make sense to do it here so we don't duplicate and have another dep.

ponderingdemocritus:
lgtm - fix merge conflict

PR #1328, opened 2024-08-20
small balancing change + use correct build cost for resource buildings

cwastche:


aymericdelab (commented):
can you just remove comments ? then all good

PR #1397, opened 2024-08-21
fix: restrain double click on battle on worldmap + small qol

edisontim:


client/src/three/scenes/Worldmap.ts
```diff
@@ -183,7 +182,10 @@ export default class WorldmapScene extends HexagonScene {
   }
 
   protected onHexagonDoubleClick(hexCoords: HexPosition) {
-    const url = new Position({ x: hexCoords.col, y: hexCoords.row }).toHexLocationUrl();
+    const position = new Position({ x: hexCoords.col, y: hexCoords.row });
+    const isBattle = this.battleManager.battles.has(undefined, position);
```
cwastche, line 186: It seems that battles.has is only called with either ID or POS, never both.
Would it be better formatting to do something like this ?
```
has(entityId: ID): boolean;
  has(position: Position): boolean;
  has(arg: ID | Position): boolean {
    if (arg instanceof Position) {
      return Array.from(this.battles.values()).some((battle) => {
        const battlePosition = battle.position.getContract();
        const positionContract = arg.getContract();
        return battlePosition.x === positionContract.x && battlePosition.y === positionContract.y;
      });
    } else {
      return this.battles.has(arg);
    }
  }
  ```
  Making the method callable without having to pass it undefined.
  
  Another option could be to divide this method in 2:
hasByPosition and hasByEntityId

aymericdelab: maybe split into 2 different methods?

aymericdelab (approved):

PR #1402, opened 2024-08-21
contracts: slash army if it leaves battle early

### LINKED ISSUE #1367: [CONTRACTS] If leaving a battle before the end slash a part of the troops

credence0x:
resolves #1367


contracts/src/systems/combat/tests/army_create_test.cairo
```diff
@@ -47,6 +47,8 @@ fn set_configurations(world: IWorldDispatcher) {
                 pillage_health_divisor: 8,
                 army_free_per_structure: 3,
                 army_extra_per_building: 2,
+                battle_leave_slash_num: 25,
+                battle_leave_slash_denom: 100
```
cwastche, line 51: Use the set_combat_config utility function for easier mods within tests

edisontim:
I've noticed that we sometimes end up with decimal points in army units. Do you think we could round the units everywhere when we divide (on battle leave, etc.) ? Otherwise we need to handle unit with < 1 troop in the client (real units)

credence0x:
> I've noticed that we sometimes end up with decimal points in army units. Do you think we could round the units everywhere when we divide (on battle leave, etc.) ? Otherwise we need to handle unit with < 1 troop in the client (real units)

@edisontim i'll look into it in a separate pr. i'll make an issue now

edisontim:
> > I've noticed that we sometimes end up with decimal points in army units. Do you think we could round the units everywhere when we divide (on battle leave, etc.) ? Otherwise we need to handle unit with < 1 troop in the client (real units)
> 
> @edisontim i'll look into it in a separate pr. i'll make an issue now

alright sounds good

edisontim (commented):
just need to address @cwastche's comment, otherwise lgtm

contracts/src/systems/combat/contracts.cairo
```diff
@@ -1547,6 +1567,20 @@ mod combat_systems {
                 battle.attack_army_health = battle_army_health;
             }
 
+            // normalize troop counts to nearest mutiple of RESOURCE_PRECISION
+            army.troops.normalize_counts();
```
edisontim, line 1571: @credence0x could you check if the order of these lines is correct?

cwastche (approved):

PR #1484, opened 2024-09-06
Disable buttons + add timer for immunity

cwastche:


aymericdelab:
add 24 ticks in grace period in EternumGlobalConfig

aymericdelab (approved):

PR #1565, opened 2024-09-12
fix: army rounding related issues

edisontim:


aymericdelab:
pnpm build failed @edisontim

aymericdelab (approved):

PR #1619, opened 2024-09-17
contracts: slash army if it leaves before battle starts

credence0x:


contracts/src/systems/combat/contracts.cairo
```diff
@@ -591,11 +591,14 @@ mod combat_systems {
             let mut attacking_army: Army = get!(world, attacking_army_id, Army);
             attacking_army.assert_not_in_battle();
 
+            println!("\n A \n\n");
```
edisontim, line 594: Remove all `println!`

PR #1662, opened 2024-09-18
Fixes update streaming for newly built buildings

cwastche:


aymericdelab:
can you try if the built/destroy works well ? @cwastche

cwastche:
> can you try if the built/destroy works well ? @cwastche

Fixed with latest commit

aymericdelab (approved):

PR #1693, opened 2024-09-20
fix: battle issues

edisontim:


client/src/dojo/modelManager/BattleManager.ts
```diff
@@ -160,10 +160,6 @@ export class BattleManager {
   public isBattleOngoing(currentTimestamp: number): boolean {
     const battle = this.getBattle();
 
-    if (this.isSiege(currentTimestamp)) {
-      return false;
-    }
-
```
cwastche, line 166: Some components were using 'isBattleOngoing' to to check its siege status.

cwastche (approved):

Issue #1719, opened 2024-09-26
[contracts + Client] Upgrading Realm


ponderingdemocritus:
**Upgrades**

Realms can only build on the immediate 6 hexs around the Castle to begin with. They must upgrade to expand the buildable area.

**There 3 levels**:
**Level 1 - ** low tier resource costs
**Level 2 - ** mid tier resources costs
**Level 3 - ** high tier resources costs

**TODO:**

- [ ] Restrict building according to the level at contract level
- [ ] Add level config setup function
- [ ] Update function to upgrade Realm
- [ ] Restrict building at client level

PR #1721, opened 2024-09-27
contracts: feature: introduce realm levels

credence0x:
contracts for #1719
- introduce realm levels
- create a function for upgrading realm levels as well as functions for configuration
- restrict building on certain hexes depending on realm level
- tests


contracts/src/models/realm.cairo
```diff
@@ -21,15 +21,16 @@ pub struct Realm {
...
 
-trait RealmCustomTrait {
-    fn has_resource(self: Realm, resource_type: u8) -> bool;
-    fn assert_is_set(self: Realm);
-}
-
+#[generate_trait]
 impl RealmCustomImpl of RealmCustomTrait {
+    fn max_level(self: Realm) -> u8 {
+        3
```
ponderingdemocritus, line 31: change this to config

ponderingdemocritus (approved):
smol config change

ponderingdemocritus:
In config i have added some contants which can be used to init the config, can you update to use them

contracts/src/models/realm.cairo
credence0x: updated

Issue #1720, opened 2024-09-26
[contracts + config] Increase Costs per building type


ponderingdemocritus:
Each building costs more than the last

**eg:**
Farm 1 = 100 Wheat
Farm 2 = 150 Wheat
Farm 3 = 200 Wheat

Keep buiulding counter * scale figure = fixed resource cost for the next building

**TODO:**
- [ ] Config 
- [ ] Building counter
- [x] Client work to adjust config so reflects around the app

cwastche:
Finished by #1727

PR #1722, opened 2024-09-27
contracts & client config: scale building cost

credence0x:
contracts and client config for #1720
- increase fixed cost of building by a certain percentage based on the number of the building type already on the realm


contracts/src/models/buildings.cairo
```diff
@@ -598,17 +583,33 @@ impl BuildingCustomImpl of BuildingCustomTrait {
         destroyed_building_category
     }
 
-    fn make_payment(world: IWorldDispatcher, entity_id: ID, category: BuildingCategory, resource_type: u8) {
-        let building_config: BuildingConfig = BuildingConfigCustomImpl::get(world, category, resource_type);
+    fn pay_fixed_cost(self: Building, building_quantity: BuildingQuantityv2, world: IWorldDispatcher) {
```
edisontim, line 586: can this keep the same name as before? I feel like the function actually pays a variable cost

contracts/src/models/config.cairo
```diff
@@ -344,6 +344,16 @@ pub struct BankConfig {
     lp_fee_denom: u128,
 }
 
+#[derive(Copy, Drop, Serde)]
+#[dojo::model]
+pub struct BuildingGeneralConfig {
+    #[key]
+    config_id: ID,
+    // cost scale percent
+    base_cost_percent_increase: u16,
```
edisontim, line 353: remove comment and change the name to be more explicit, we don't want comments to be everywhere if it can be explained through the name imo

credence0x: i dont actually know what a better name would be. ive renamed it like 5 times

edisontim: `increase_per_extra_building_percent` ? Honestly I prefer explicit and long than short haha

edisontim: `quadratic_increase_per_extra_building_percent`

credence0x: it doesnt fit into a felt. long names wont fit into a felt

edisontim: Forgot dojo had those limitations

credence0x: 31 characters at most

edisontim (approved):
lgtm

PR #1741, opened 2024-10-05
update battle time calculations

credence0x:
make the battle time calculations more sane
- there's now a config to increase and reduce battle time more directly through `battle_time_scale`
- battle time goes up with the number of troops in battle and there is no upper time limit
- the smaller army now loses troops quicker than the bigger army

### DIFF
--- a/contracts/src/utils/math.cairo
+++ b/contracts/src/utils/math.cairo
@@ -30,6 +30,16 @@ pub fn max<T, impl TPartialOrd: PartialOrd<T>, impl TCopy: Copy<T>, impl TDrop: 
     };
 }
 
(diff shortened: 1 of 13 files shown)


contracts/src/utils/math.cairo
```diff
@@ -30,6 +30,16 @@ pub fn max<T, impl TPartialOrd: PartialOrd<T>, impl TCopy: Copy<T>, impl TDrop:
     };
 }
 
+pub fn cap_minus<T, impl TPartialOrd: PartialOrd<T>, impl TSub: Sub<T>, impl TCopy: Copy<T>, impl TDrop: Drop<T>>(
+    a: T, b: T
+) -> T {
+    return if (a < b) {
+        return a - a; // 0
```
edisontim, line 37: why not just return 0 in this case?

edisontim (changes requested):
Could you provide a fomula that describes the battle_time based on input troops? Also a concrete case would be helpful

PR #1857, opened 2024-10-16
fix: resource prod fragment mines

edisontim:


client/src/ui/components/resources/ResourceChip.tsx
```diff
@@ -28,13 +35,27 @@ export const ResourceChip = ({
...
+  const [balance, setBalance] = useState(productionManager.balance(currentDefaultTick));
+
   const production = useMemo(() => {
     return productionManager.getProduction();
   }, [productionManager]);
 
-  const balance = useMemo(() => {
-    return productionManager.balance(currentDefaultTick);
-  }, [productionManager, production, currentDefaultTick, maxStorehouseCapacityKg]);
+  useEffect(() => {
```
ponderingdemocritus, line 44: have you checked this in prod? the useEffect here might be savage on renders

PR #1861, opened 2024-10-17
feat: randomize mercenaries number

### LINKED ISSUE #1844: feat: randomise mercenaries between x amounts
instead of 1000 mercs on objects, lets make it random

It should exist from 1000-5000 randomly on locations. This will create some nice drama and also unpredictability.

edisontim:
closes #1844

### DIFF
--- a/contracts/src/systems/config/contracts.cairo
+++ b/contracts/src/systems/config/contracts.cairo
@@ -150,7 +150,16 @@ trait IPopulationConfig {
 
 #[dojo::interface]
 trait IMercenariesConfig {
-    fn set_mercenaries_config(ref world: IWorldDispatcher, troops: Troops, rewards: Span<(u8, u128)>);
+    fn set_mercenaries_config(
+        ref world: IWorldDispatcher,
+        knights_lower_bound: u64,
+        knights_upper_bound: u64,
+        paladins_lower_bound: u64,
+        paladins_upper_bound: u64,
+        crossbowmen_lower_bound: u64,
+        crossbowmen_upper_bound: u64,
+        rewards: Span<(u8, u128)>
+    );
 }
 
 #[dojo::interface]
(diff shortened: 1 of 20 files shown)


contracts/src/systems/config/contracts.cairo
```diff
@@ -150,7 +150,16 @@ trait IPopulationConfig {
 
 #[dojo::interface]
 trait IMercenariesConfig {
-    fn set_mercenaries_config(ref world: IWorldDispatcher, troops: Troops, rewards: Span<(u8, u128)>);
+    fn set_mercenaries_config(
+        ref world: IWorldDispatcher,
+        knights_lower_bound: u64,
```
ponderingdemocritus, line 155: we could make these move compressed no? would this fit in u32 or smaller?

ponderingdemocritus (commented):
minor suggestion

sdk/packages/eternum/src/constants/global.ts
```diff
@@ -95,9 +95,12 @@ export const TROOP_BATTLE_LEAVE_SLASH_DENOM = 100;
 export const TROOP_BATTLE_TIME_REDUCTION_SCALE = 1_000;
 
 // Mercenaries
-export const MERCENARIES_KNIGHT_COUNT = 1000;
-export const MERCENARIES_PALADIN_COUNT = 1000;
-export const MERCENARIES_CROSSBOWMAN_COUNT = 1000;
+export const MERCENARIES_KNIGHTS_LOWER_BOUND = 1000;
+export const MERCENARIES_KNIGHTS_UPPER_BOUND = 5000;
+export const MERCENARIES_PALADINS_LOWER_BOUND = 1000;
+export const MERCENARIES_PALADINS_UPPER_BOUND = 5000;
```
aymericdelab, line 101: increase higher bound to 10k for each troop type

aymericdelab:
@edisontim you need to change set_mercenaries_config in the tests

contracts/src/systems/config/contracts.cairo
edisontim: hmmm the knights_count, paladin_count and crossbowman_count inside the Troops structs are all u64
Also it would make the bounds max 4,294,967 troops accounting for the 1000 precision needed

PR #1862, opened 2024-10-17
[contracts + client] season pass

credence0x:
- created a season pass contract as well as a test realms contract
- get realm metadata from contracts rather than passing it through client. now, realms can no longer be duplicated
- remove unnecessary fields from Realm model
- update client to use season pass
- formula optimizations here and there


contracts/src/constants.cairo
```diff
@@ -75,28 +75,29 @@ fn all_resource_ids() -> Array<u8> {
 
 // Note: Please update the all_resources_ids list whenever ResourceTypes are updated
 mod ResourceTypes {
-    const WOOD: u8 = 1;
```
ponderingdemocritus, line 78: why have these changed?

sdk/packages/eternum/src/constants/index.ts
```diff
@@ -12,27 +12,27 @@ export * from "./troops";
 export * from "./utils";
 
 export enum ResourcesIds {
-  Wood = 1,
-  Stone = 2,
-  Coal = 3,
+  Stone = 1,
+  Coal = 2,
+  Wood = 3,
```
ponderingdemocritus, line 17: why did ids change?

season_pass/contracts/src/contract.cairo
```diff
@@ -0,0 +1,134 @@
...
+// Compatible with OpenZeppelin Contracts for Cairo ^0.17.0
+
+// Eternum Season Pass
+#[starknet::interface]
+trait IRealmMetadataEncoded<TState> {
+    fn get_encoded_metadata(self: @TState, token_id: u16) -> (felt252, felt252, felt252);
+}
+
+#[starknet::contract]
+mod EternumSeasonPass {
```
ponderingdemocritus, line 11: Where are we going to check that the pass has been used? I think for sanity, we should burn the pass if it is used.

credence0x: used passes would be owned by the realms system contract. we transfer the pass to the realms system contract. we could also burn

ponderingdemocritus (commented):
Looks good - confused why resource IDS have changed

season_pass/contracts/src/contract.cairo
ponderingdemocritus: Ok - let's leave that. Then we can track as we go.

PR #1909, opened 2024-10-25
[client] show more context in the market modal

### LINKED ISSUE #1886: [client] notification when AMM get's taken
show somewhere in the UI either that:
- bank will be taken in X hours and will be unavailable
- bank is now taken, will be available in X hours

maybe on the trade module button on the left navigation ?

aymericdelab:
Fixes #1886


client/src/ui/components/trading/MarketModal.tsx
```diff
@@ -59,6 +62,21 @@ export const MarketModal = () => {
...
 
+  const battles = useBattlesByPosition(bank?.position || { x: 0, y: 0 });
+
+  const currentBlockTimestamp = useUIStore.getState().nextBlockTimestamp || 0;
+
+  const getStructure = useStructureByPosition();
+
+  const bankStructure = getStructure(bank?.position || { x: 0, y: 0 });
+
+  const battle = useMemo(() => {
```
edisontim, line 73: this will not be the bank battle if another battle between two armies happen on the same hex I think

edisontim (changes requested):

ponderingdemocritus:
lgtm - just address comment

PR #1926, opened 2024-10-30
feat: add sounds

### LINKED ISSUE #1919: [client] add sounds
- hex selection in building mode/ travel mode
- when creating/accepting orders/ swapping
- when clicking on a building

### LINKED ISSUE #1924: [client] add screaming donkey sound when burning donkeys

edisontim:
Closes #1919 
Closes #1924


client/src/three/components/BuildingPreview.ts
```diff
@@ -86,6 +88,11 @@ export class BuildingPreview {
 
   public setBuildingPosition(position: THREE.Vector3) {
     if (this.previewBuilding) {
+      if (!this.currentHexHovered || !this.currentHexHovered.equals(position)) {
+        new Audio(dir + soundSelector.hoverClick).play();
```
r0man1337, line 92: There was `useShovelSound()` hook in useUiSounds file, that plays two similar sounds one by one.
This sound was designed exactly for hex selection in building mode.

May be we can reuse it here?

aymericdelab: yeah let's do it

client/src/three/scenes/Hexception.tsx
```diff
@@ -629,3 +650,29 @@ export default class HexceptionScene extends HexagonScene {
     });
   }
 }
+
+export const playBuildingSound = (buildingType: BuildingType | undefined) => {
```
aymericdelab, line 654: shouldn't all this logic be in the useUISound file ?

client/src/three/components/BuildingPreview.ts
edisontim: Should I change it for the hex selection in the worldmap view as well?

PR #1927, opened 2024-10-31
Bob/dev

bob0005:
Merge leaderboard & guild component


client/src/types/guild.ts
```diff
@@ -0,0 +1,44 @@
+import { ContractAddress, ID } from "@bibliothecadao/eternum";
+
+export type GuildAndName = {
```
aymericdelab, line 3: why do you need so many different types for guild ? would be more clear if you could have one GuildInfo type. Also move it to sdk

client/src/hooks/helpers/useGuilds.tsx
```diff
@@ -100,23 +81,41 @@ export const useGuilds = () => {
 
   const useGuildMembers = useCallback((guildEntityId: ID) => {
     const guildMembers = useEntityQuery([HasValue(GuildMember, { guild_entity_id: guildEntityId })]);
```
aymericdelab, line 83: don't use hook inside useCallback

client/src/hooks/helpers/useGuilds.tsx
```diff
@@ -100,23 +81,41 @@ export const useGuilds = () => {
...
+        getAddressName,
+        GuildMember,
+        Owner,
+        JoinGuild,
+      ),
     };
   }, []);
 
   const useGuildQuery = useCallback(() => {
     const guilds = useEntityQuery([Has(Guild), NotValue(Guild, { member_count: 0 })]);
```
aymericdelab, line 99: don't use hook inside useCallback

client/src/ui/components/navigation/OSWindow.tsx
```diff
@@ -1,7 +1,47 @@
+import { ExpandablePopup } from "@/ui/elements/ExpandablePopup";
 import { SecondaryPopup } from "../../elements/SecondaryPopup";
 import { OSInterface } from "./Config";
 
-export const OSWindow = ({ onClick, show, title, children, width = "400px", hintSection }: OSInterface) => {
+interface ExpandableOSInterface extends OSInterface {
```
aymericdelab, line 5: tried expanding the social window but didn't work

client/src/ui/components/worldmap/players/PlayersPanel.tsx
```diff
@@ -39,44 +54,196 @@ export const PlayersPanel = () => {
...
+        isInvited =
+          getComponentValue(GuildWhitelist, getEntityIdFromKeys([player.address, BigInt(userGuild?.guildEntityId)]))
+            ?.is_whitelisted ?? false;
+      }
       return {
         name: player.addressName,
         address: player.address,
         structures,
+        isUser: player.address === ContractAddress(account.address),
+        rank: "#" + (index != -1 ? String(index + 1) : "NA"),
```
aymericdelab, line 69: don't put NA , just rank randomly if two people have the same points

client/src/ui/elements/ExpandablePopup.tsx
```diff
@@ -0,0 +1,59 @@
+import { AnimatePresence, motion } from "framer-motion";
+import React from "react";
+import { SecondaryPopup } from "../elements/SecondaryPopup";
+
+interface ExpandablePopupProps {
```
aymericdelab, line 5: think it would be better to put ExpandablePopup in the same place as the OSWindow, and have 2 different components, ExandableOSWindow and OSWindow, seems weird to have one component OSWindow with condition returning either one component or the other

contracts/src/models/event.cairo
```diff
@@ -158,3 +158,23 @@ pub struct SettleRealmData {
     y: u32,
     timestamp: u64,
 }
+
+#[derive(Copy, Drop, Serde)]
+#[dojo::event]
+#[dojo::model]
+pub struct CreateGuild {
```
aymericdelab, line 165: could you add guild name in both events

PR #1933, opened 2024-11-02
🏆 Integrate controller achievements

bal7hazar:
Som additional context:
- `eternum::utils::tasks`: contains all individual unit tasks and corresponding metadata
- `eternum::utils::trophies`: contains all individual achievements/trophies and corresponding metadata, each trophy also describe a list of `Task` to be achieved
- `eternum::systems::config::contracts::config_systems::dojo_init`: leveraging the dojo constructor to emit all events corresponding to the Trophy definition (so they are emitted only once)
- `eternum::systems::*`: at different location I have emitted a task progression, these will need a triple check to ensure they are at the right position in the flow. Especially for the **Battlelord** one ([source](https://github.com/bal7hazar/eternum/blob/9df88dd16a8a8fab2b13eeab86b53e6ad8b441f0/contracts/src/systems/combat/contracts/battle_systems.cairo#L686)) which should be emitted when a battle is won (it will be done for each army leave a winning battle, but I dont think we could do better), I checked that the battle is correctly over and the army.count is > 0 not sure if it was enough.

Note: there are two way to manage these events, I started with the native starknet components that I have implemented in the contracts when needed, but I was facing an issue for the donkey burn achievement where the ContractState was not available and so hard to implement components, so I did a lower level management of this event (based only on a WorldStorage), I still feel like the component way should be the standard so I left it like this, but technically speaking we could use the low level approach everywhere and avoid some extra code to implement components everywhere.


contracts/src/systems/transport/contracts/donkey_systems.cairo
```diff
@@ -46,15 +48,26 @@ mod donkey_systems {
...
                         player_address: starknet::get_caller_address(),
                         amount: donkey_amount,
-                        timestamp: starknet::get_block_timestamp()
+                        timestamp: time
                     }
                 );
+
+            // [Achievement] Consume donkeys
+            if donkey_amount != 0 {
+                let count = (donkey_amount % core::integer::BoundedU32::max().into()).try_into().unwrap();
```
bal7hazar, line 64: I would like to have an integer rounded down here (without the `decimals` precision)
I am not sure were to fetch this decimal from here, if you have an idea @credence0x

contracts/src/systems/combat/contracts/battle_systems.cairo
```diff
@@ -658,6 +683,11 @@ mod battle_systems {
                 world.write_model(@army);
                 world.write_model(@army_health);
                 world.write_model(@army_quantity);
+            } else if army.troops.count() > 0 {
```
bal7hazar, line 686: Nice one needs a double check, the expectation is leaving a winning battle

bal7hazar (commented):
Few comments to contextualise

contracts/src/systems/combat/contracts/battle_systems.cairo
credence0x: the `BattleLord` achievement should only  be inside the internal `fn leave_battle`. inside the function, check if the army left early then give the reward if the army didnt leave early and they are still alive. it should be at the bottom of the function. then you can remove the code from everywhere else

contracts/src/systems/realm/contracts.cairo
```diff
@@ -242,6 +274,16 @@ mod realm_systems {
 
             quest.completed = true;
             world.write_model(@quest);
+
```
credence0x, line 277: could you add a check after `let quest_reward_config: QuestRewardConfig = world.read_model(quest_id);` to assert that `quest_reward_config.detached_resource_count > 0`. so we're sure that the last quest is really the last one. they cant just put an arbitrary `quest_id`

contracts/src/systems/transport/contracts/donkey_systems.cairo
credence0x: to get the count without precision, it would be `let count = donkey_amount  / RESOURCE_PRECISION;`. the precision can be gotten from `eternum::constants::RESOURCE_PRECISION`. 

i think that could lead to an innacurate donkey burn count though. if we want it to be accurate, we should use the precision

credence0x (changes requested):

contracts/src/systems/combat/contracts/battle_systems.cairo
```diff
@@ -1155,6 +1183,15 @@ pub impl InternalBattleImpl of InternalBattleTrait {
 
     /// Make army leave battle
     fn leave_battle(ref world: WorldStorage, ref battle: Battle, ref original_army: Army) {
+        // [Achievement] Win a battle
+        if battle.has_ended() && original_army.troops.count() > 0 {
```
credence0x, line 1187: you should make a variable at the top to know if battle ended, then use the updated army at the bottom when checking if army count is > 0

credence0x (changes requested):

credence0x (approved):
lgtm

PR #2116, opened 2024-11-15
fix: stop counting the points when a season has ended

edisontim:


client/src/hooks/store/useLeaderBoardStore.tsx
```diff
@@ -75,4 +76,16 @@ export const useSubscriptionToHyperstructureEvents = () => {
 
     return () => subscription.unsubscribe();
   }, []);
+
+  const gameEndedEntityId = useEntityQuery([Has(GameEnded)]);
```
aymericdelab, line 80: why not use useComponentValue directly ? we already know the entity id of the GameEnded entity

aymericdelab, line 80: why not use useComponentValue directly ? we already know the entity id of the GameEnded entity

aymericdelab: oh didnt realize winner_address was a key (which isnt needed imo) but let's keep it like this then, lgtm

PR #2130, opened 2024-11-20
[client] Use production everywhere

### LINKED ISSUE #1080: [client] Use production everywhere
If we rename production manager to ResourceManager, we use a principle of Clean Code (one interface for all things similar) to only use this interface for all places where we use resources
The production manager doesn't cause any re-renders in case the resource isn't being produced

aymericdelab:
Fixes #1080

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `ResourceManager` class for enhanced resource management.
	- Added a method to specifically fetch resource information.

- **Refactor**
	- Renamed and replaced instances of `ProductionManager` with `ResourceManager` across various components and hooks.
	- Updated related methods to align with the new resource management approach.

- **Bug Fixes**
	- Adjusted logic in components to ensure proper resource management functionality remains intact.

### DIFF
--- a/client/src/hooks/helpers/useResources.tsx
+++ b/client/src/hooks/helpers/useResources.tsx
@@ -25,13 +25,15 @@ export function getResourcesUtils() {
   } = useDojo();
 
   const getResourcesFromBalance = (entityId: ID): Resource[] => {
+    const currentDefaultTick = useUIStore.getState().currentDefaultTick;
+
     const ownedResources = getComponentValue(OwnedResourcesTracker, getEntityIdFromKeys([BigInt(entityId)]));
     if (!ownedResources) return [];
     const resourceIds = resources.map((r) => r.id);
     return resourceIds
       .map((id) => {
         const resourceManager = new ResourceManager(useDojo().setup, entityId, id);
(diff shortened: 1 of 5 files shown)


client/src/hooks/helpers/useResources.tsx
```diff
@@ -12,28 +12,29 @@ import {
...
+
     const ownedResources = getComponentValue(OwnedResourcesTracker, getEntityIdFromKeys([BigInt(entityId)]));
     if (!ownedResources) return [];
-    // const resourceIds = getResourceIdsFromPackedNumber(ownedResources.resource_types);
     const resourceIds = resources.map((r) => r.id);
     return resourceIds
       .map((id) => {
-        const resource = getComponentValue(Resource, getEntityIdFromKeys([BigInt(entityId), BigInt(id)]));
-        return { resourceId: id, amount: Number(resource?.balance) || 0 };
+        const resourceManager = new ResourceManager(useDojo().setup, entityId, id);
```
bob0005, line 35: We should remove getResourceCosts() as it isn't used anymore and has been replaced by a configManager method.
Also, is there a specific need to call useDojo here again instead of passing it from the first call at the top of function ?

bob0005 (approved):
lgtm, small nit comment

PR #2136, opened 2024-11-21
Fix building destroy + remove army model on troop exchange

### LINKED ISSUE #1750: [Client] Transferring all troops from an army to another doesn't remove its 3d model on the map

bob0005:
Closes #1750

### DIFF
--- a/contracts/src/models/buildings.cairo
+++ b/contracts/src/models/buildings.cairo
@@ -260,8 +260,6 @@ impl BuildingProductionCustomImpl of BuildingProductionCustomTrait {
             // take back bonuses given to surrounding buildings if this building is a bonus supplier
             self.update_bonuses_supplied(ref world, delete: true);
         }
-
-        world.write_model(@self);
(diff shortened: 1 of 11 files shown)


contracts/src/models/buildings.cairo
```diff
@@ -260,8 +260,6 @@ impl BuildingProductionCustomImpl of BuildingProductionCustomTrait {
             // take back bonuses given to surrounding buildings if this building is a bonus supplier
             self.update_bonuses_supplied(ref world, delete: true);
         }
-
-        world.write_model(@self);
     }
```
aymericdelab, line 263: can you confirm it's ok to remove this?

bob0005: Yes. This is what was causing issues when trying to destroy a building.
The model is already being written in the calling contract after this internal call, and pausing buildings still works as intended.

edisontim (approved):

PR #2265, opened 2024-12-07
contracts: remove bank creation system

credence0x:


ponderingdemocritus:
we will be using this in future season - we should just disable

PR #2280, opened 2024-12-07
Add variable hyperstructure construction cost

credence0x:
## Release Notes

- **New Features**
	- Introduced a `randomness` property to the `Hyperstructure` component, enhancing its data model.
	- Added new methods for calculating hyperstructure resource requirements, incorporating tiered resource management.
	- Implemented a new `HyperstructureCompletionTable` for improved display of completion costs.

- **Bug Fixes**
	- Enhanced point calculation logic in the leaderboard to account for hyperstructure entity IDs.

- **Documentation**
	- Updated configuration setup logic to include error handling for environment variables.

- **Refactor**
	- Revised resource cost structures to use tiered representations with minimum and maximum amounts.

- **Chores**
	- Improved error handling and user prompts in the configuration setup process.

### DIFF
--- a/sdk/packages/eternum/src/constants/structures.ts
+++ b/sdk/packages/eternum/src/constants/structures.ts
@@ -1,7 +1,6 @@
 import { CairoCustomEnum } from "starknet";
 import { ResourcesIds } from "../constants";
 import { ResourceInputs } from "../types";
-import { HYPERSTRUCTURE_CREATION_COSTS } from "./hyperstructure";
 
 // Knip ignore tag
 /** @public */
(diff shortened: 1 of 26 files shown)


sdk/packages/eternum/src/constants/structures.ts
```diff
@@ -180,7 +179,7 @@ export function determineEntityState(
 }
 
 export const STRUCTURE_COSTS: ResourceInputs = {
-  [StructureType.Hyperstructure]: HYPERSTRUCTURE_CREATION_COSTS,
+  [StructureType.Hyperstructure]: [], //todo hyperstructure costs
```
ponderingdemocritus, line 182: todo?

credence0x: @aymericdelab does that need to be updated? i just removed it so it would compile

PR #2610, opened 2024-12-27
Fix build

bob0005:


edisontim (commented):
This doesn't mess up anything? Isn't the point to have the rewards popup open directly?

aymericdelab (approved):

PR #2660, opened 2025-01-23
client: make client biome impl equivalent to contracts

credence0x:
- make client biome implementation equivalent to contracts by using fixed point math in client. it was previously not equivalent because of precision issues in many places

## Release Notes

- **New Features**
	- Introduced advanced fixed-point arithmetic system for precise mathematical calculations.
	- Enhanced biome generation with improved noise generation algorithms.
	- Added new classes for three-dimensional (`Vec3`) and four-dimensional (`Vec4`) vector operations.

- **Technical Improvements**
	- Implemented comprehensive vector and mathematical utility classes.
	- Upgraded computational precision for terrain and biome generation.
	- Added support for fixed-point arithmetic operations and enhanced type safety in biome calculations.

- **Performance**
	- Optimized noise generation and vector computation methods.
	- Improved mathematical precision in terrain generation algorithms.

### DIFF
--- a/client/apps/game/src/three/managers/biome.ts
+++ b/client/apps/game/src/three/managers/biome.ts
@@ -1,10 +1,17 @@
-import { snoise } from "@dojoengine/utils";
+import { Fixed, FixedTrait } from "@/utils/biome/fixed-point";
+import { noise as snoise } from "@/utils/biome/simplex-noise";
+import { Vec3 } from "@/utils/biome/vec3";
 import * as THREE from "three";
 
-const MAP_AMPLITUDE = 60;
-const MOISTURE_OCTAVE = 2;
-const ELEVATION_OCTAVES = [1, 0.25, 0.1];
-const ELEVATION_OCTAVES_SUM = ELEVATION_OCTAVES.reduce((a, b) => a + b, 0);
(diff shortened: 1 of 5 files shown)


client/apps/game/src/three/managers/biome.ts
```diff
@@ -1,10 +1,17 @@
...
+const MOISTURE_OCTAVE = FixedTrait.fromInt(2n);
+const ELEVATION_OCTAVES = [
+  FixedTrait.fromInt(1n),  // 1
+  FixedTrait.divi(FixedTrait.fromInt(1n), FixedTrait.fromInt(4n)),  // 0.25
+  FixedTrait.divi(FixedTrait.fromInt(1n), FixedTrait.fromInt(10n))  // 0.1
+];
+const ELEVATION_OCTAVES_SUM = ELEVATION_OCTAVES.reduce((a, b) => a.add(b), FixedTrait.ZERO);
+
 
 export type BiomeType =
```
aymericdelab, line 16: should we use an enum instead ?

aymericdelab (approved):

client/apps/game/src/three/managers/biome.ts
credence0x: updated

PR #2663, opened 2025-01-23
feat: add Torii local launcher

edisontim:
Allows a user to simply download a .dmg file built from this app, install the dmg and then run a local torii pointing to mainnet.

Configuration is dynamic, user can choose to use a custom RPC, mainnet, sepolia or localhost.

Further improvement, handle torii version if torii is already installed on user's machine. Currently it checks if the user has torii and in that case just uses the version of the user. Otherwise it downloads the compatible version

## Summary by CodeRabbit

- New Features  
  • Refreshed launcher interface featuring interactive components such as customizable settings for network endpoints, real-time synchronization progress, warning prompts, and toast notifications that enhance user engagement.

- Chores  
  • Upgraded build tools and configuration settings, including dependency updates, to improve reliability and performance behind the scenes.


aymericdelab:
can you only keep public folder and remove assets ? 

image

aymericdelab:
can you use kebab case for all files ?

PR #3470, opened 2025-06-25
blitz mode

credence0x:


contracts/game/src/models/resource/resource.cairo
```diff
@@ -128,9 +129,24 @@ pub impl TroopResourceImpl of TroopResourceTrait {
 
 #[generate_trait]
 pub impl SingleResourceImpl of SingleResourceTrait {
+    fn is_relic_or_essence(resource_type: u8) -> bool {
+        (resource_type >= RELICS_RESOURCE_START_ID && resource_type <= RELICS_RESOURCE_END_ID)
+            || resource_type == ResourceTypes::ESSENCE
+    }
+
+    fn ensure_correct_precision(ref self: SingleResource) {
```
aymericdelab, line 137: to me essence is like another resource, so why not precision ?

PR #3679, opened 2025-09-28
fix of 3669

ponderingdemocritus:


ponderingdemocritus:
@credence0x can you review this as it's fixing client/contract drift

aymericdelab (approved):

PR #4250, opened 2026-02-19
fix: skip stamina/cooldown check when claiming unguarded structures

kingofrealms69:
Fixes the quick attack preview to allow claiming unguarded structures without stamina or cooldown requirements.

- Skip stamina requirement and cooldown for unguarded structure claims
- Show "Claim" button label instead of "Attack" when no defenders
- Prioritize "No troops selected" check first

Clean version of #4241 (removed unrelated onchain-agent changes).


credence0x (approved):
lgtm

PR #4571, opened 2026-04-10
Add YAML-backed game launch schedules

ponderingdemocritus:
This adds a minimal YAML-backed path for launching scheduled games through the existing clean deployer workflow. The goal is to keep the current GitHub Actions and run-store flow intact while letting weekly schedules live in a committed config file instead of inline dispatch JSON.

What changed:
- add `--config-path` support to the clean deployer request builders so game, series, and rotation launches can be loaded from YAML
- add a small helper CLI that turns a config-backed launch request into the `GAME_LAUNCH_*` env exports the current workflow already expects
- update `.github/workflows/game-launch.yml` to accept `config_path` and hydrate launch inputs from YAML before the existing step runner takes over
- document the new path and add an example weekly series config under `config/deployer/clean/examples/`

Why this shape:
- recovery, auto-retry, and run-store persistence already operate on the typed launch request objects, so the least risky change is to normalize YAML into those existing request types
- GitHub still needs `launch_kind`, `environment`, and `launch_step` as dispatch inputs because the workflow routes on them before the shell scripts run

Verification:
- `bun test config/deployer/clean/tests/launch-request.test.ts`
- `bun test config/deployer/clean/tests/cloudflare-worker.test.ts`
- `pnpm run format`
- `pnpm run knip` *(fails in this workspace because `client/apps/eternum-mobile/eslint.config.js` cannot resolve `@eslint/js`)*


config/deployer/clean/cli/launch-workflow-env.ts
```diff
@@ -0,0 +1,136 @@
...
+  assignOptionalLaunchOption(launchOptions, "verboseConfigLogs", request.verboseConfigLogs);
+  assignOptionalLaunchOption(launchOptions, "version", request.version);
+  assignOptionalLaunchOption(launchOptions, "maxActions", request.maxActions);
+  assignOptionalLaunchOption(launchOptions, "waitForFactoryIndexTimeoutMs", request.waitForFactoryIndexTimeoutMs);
+  assignOptionalLaunchOption(launchOptions, "waitForFactoryIndexPollMs", request.waitForFactoryIndexPollMs);
+  assignOptionalLaunchOption(launchOptions, "skipIndexer", request.skipIndexer);
+  assignOptionalLaunchOption(launchOptions, "skipLootChestRoleGrant", request.skipLootChestRoleGrant);
+  assignOptionalLaunchOption(launchOptions, "skipBanks", request.skipBanks);
+  assignOptionalLaunchOption(launchOptions, "dryRun", request.dryRun);
+
```
ponderingdemocritus, line 51: `workflowFile` and `ref` are accepted from YAML here: https://github.com/BibliothecaDAO/eternum/blob/9406abfa844a6e4aa45680806f99d60e1baf2dc7/config/deployer/clean/cli/launch-config-file.ts#L7-L19 and carried onto the request here: https://github.com/BibliothecaDAO/eternum/blob/9406abfa844a6e4aa45680806f99d60e1baf2dc7/config/deployer/clean/cli/launch-request.ts#L350-L395 . They are consumed by indexer dispatch to choose the workflow target/ref, but this serializer drops both fields when it converts the request back into `GAME_LAUNCH_*` exports for Actions. Result: config-backed launches silently fall back to the default workflow/current ref instead of the YAML-specified dispatch target.

.github/workflows/game-launch.yml
```diff
@@ -176,6 +181,19 @@ jobs:
...
+
+          if [ -n "${GAME_LAUNCH_CONFIG_PATH:-}" ]; then
+            bun config/deployer/clean/cli/print-launch-workflow-env.ts \
+              --config-path "${GAME_LAUNCH_CONFIG_PATH}" \
+              --launch-kind "${GAME_LAUNCH_KIND}" \
+              --environment "${GAME_LAUNCH_ENVIRONMENT}" \
+              > .context/game-launch/launch-config-inputs.sh
+
+            # shellcheck disable=SC1091
+            source .context/game-launch/launch-config-inputs.sh
```
ponderingdemocritus, line 194: This `source` call overwrites caller-supplied workflow inputs with the YAML-derived exports. That breaks the explicit precedence already implemented for config-backed launches, where CLI args are supposed to win over config values: https://github.com/BibliothecaDAO/eternum/blob/9406abfa844a6e4aa45680806f99d60e1baf2dc7/config/deployer/clean/cli/launch-config-file.ts#L391-L397 . Repro: dispatch with `config_path=...` and `launch_options_json={"durationSeconds":3600}` or `auto_retry_enabled=false`; after sourcing `launch-config-inputs.sh`, `GAME_LAUNCH_OPTIONS_JSON` / `GAME_LAUNCH_AUTO_RETRY_ENABLED` are reset to the YAML/default exports, so the operator override is silently ignored.

PR #4640, opened 2026-04-15
fix(game): pass entityId to 3D label stamina resolution

ponderingdemocritus:
## Summary

The 3D army label on the map showed different stamina values than the React UI panel for the same army. The root cause was that `resolveArmyStaminaSnapshot` in `army-manager.ts` omitted `entityId` when calling `getExplorerStaminaSnapshot()`, preventing the authoritative troops snapshot store from being consulted. This caused the 3D label to fall back to troops with zeroed boosts, losing stamina regen bonuses. The fix adds `entityId: input.entityId` to the call, aligning the 3D label with the UI panel.

A new test verifies that `selectFreshestTroopsSnapshot` returns the authoritative snapshot (with boosts) when `entityId` is provided, and falls back to zeroed boosts when it is not.


client/apps/game/src/ui/features/world/components/entities/hooks/use-army-entity-detail.ts
```diff
@@ -72,35 +68,33 @@ export const useArmyEntityDetail = ({ armyEntityId }: UseArmyEntityDetailOptions
   const explorer = explorerData?.explorer;
   const explorerResources = explorerData?.resources;
 
-  useEffect(() => {
-    if (!explorer?.troops) {
-      return;
+  // Use live RECS troops as single source of truth, fall back to Torii query snapshot
+  const troops: Troops | null = useMemo(() => {
+    const candidate = liveExplorerTroops ?? explorer?.troops ?? null;
```
ponderingdemocritus: This now always prefers `liveExplorerTroops` when present, without comparing `stamina.updated_tick` or preserving the same-tick snapshot priority from `getExplorerStaminaSnapshot`. Existing coverage catches the regression: with Node 20.19, `use-army-entity-detail.stamina-sync.test.tsx` now renders stale live values (`20/120:20`, `80/120:80`) where the tests expect the fresher/authoritative Torii snapshots (`65/120`, `20/120:30`).

client/apps/game/src/ui/features/world/components/entities/banner/army-banner-entity-detail.tsx
```diff
@@ -130,6 +130,7 @@ const ArmyBannerEntityDetailContent = memo(
               maxStamina={derivedData.maxStamina}
               displayRatio={derivedData.staminaDisplay?.displayRatio}
               isRecharging={derivedData.staminaDisplay?.isRecharging}
+              displayCurrent={derivedData.staminaDisplay?.displayCurrent}
```
ponderingdemocritus: This changes the user-facing stamina display, but the diff does not update `client/apps/game/src/ui/features/world/latest-features.ts`. The scoped client instructions require that for UX changes: https://github.com/BibliothecaDAO/eternum/blob/1a88b3d915815ba9c9ad42623d72fecd3337e46a/client/apps/game/AGENTS.md#documenting-ux-changes-and-new-features

client/apps/game/src/three/managers/army-manager.ts
```diff
@@ -2779,39 +2785,84 @@ ${
...
     }
 
-    const staminaDisplay = buildProjectedStaminaDisplayModel({
-      committedCurrent: staminaSnapshot.current,
-      committedMax: staminaSnapshot.max,
-      armiesTickTimeRemaining,
-      currentArmiesTick,
-      troops: staminaSnapshot.troops,
-    });
+    const liveTroops = this.resolveLiveExplorerTroops(input.entityId);
```
ponderingdemocritus: `resolveArmyStaminaSnapshot` no longer calls `getExplorerStaminaSnapshot` / `selectFreshestTroopsSnapshot` with `entityId`, so the authoritative per-entity snapshot path that this PR tests is not used by the production 3D label path. `rg` also shows there are no production callers of `setAuthoritativeTroopsSnapshot` anymore, so when RECS is missing or still hydrating this falls back to `buildMinimalTroops` with zeroed boosts and can still drop regen bonuses.

PR #4641, opened 2026-04-15
Stabilize production automation

ponderingdemocritus:
## Summary
- serialize per-realm production automation transactions to avoid same-signer nonce races
- update production planning so custom inactive resources are diagnosed, troop dependencies are ordered, and Smart can run one affordable troop cycle
- surface concrete skip reasons in the production automation dashboard

## Verification
- PATH="/Users/os/Library/pnpm:$PATH" pnpm --dir client/apps/game test -- automation-processor automation-runner
- PATH="/Users/os/Library/pnpm:$PATH" pnpm --dir client/apps/game typecheck
- PATH="/Users/os/Library/pnpm:$PATH" pnpm run format
- PATH="/Users/os/Library/pnpm:$PATH" pnpm run knip

### DIFF
--- a/client/apps/game/src/ui/features/infrastructure/automation/model/automation-processor.ts
+++ b/client/apps/game/src/ui/features/infrastructure/automation/model/automation-processor.ts
@@ -265,9 +265,11 @@ export const buildRealmProductionPlan = ({
     }
   });
 
-  const configuredCustomIds = Object.keys(realmConfig.customPercentages ?? {}).map(
-    (key) => Number(key) as ResourcesIds,
-  );
+  const presetId = realmConfig.presetId ?? "smart";
+  const configuredCustomIds =
+    presetId === "custom"
+      ? Object.keys(realmConfig.customPercentages ?? {}).map((key) => Number(key) as ResourcesIds)
+      : [];
   const resourceIdsToEvaluate = buildResourceIdsToEvaluate(producedResourceIds, configuredCustomIds);
 
   if (resourceIdsToEvaluate.length === 0) {
@@ -285,7 +287,6 @@ export const buildRealmProductionPlan = ({
     };
   }
 
-  const presetId = realmConfig.presetId ?? "smart";
   const smartDefaults = calculatePresetAllocations(resourceIdsToEvaluate, "smart", entityType);
   const presetAllocations =
     presetId === "smart" || presetId === "idle"
--- a/client/apps/game/src/ui/features/infrastructure/automation/model/automation-processor.ts
+++ b/client/apps/game/src/ui/features/infrastructure/automation/model/automation-processor.ts
@@ -115,6 +115,86 @@ const compareAutomationResources = (left: ResourcesIds, right: ResourcesIds): nu
 const isResourceActiveInSnapshot = (snapshot: RealmResourceSnapshot, resourceId: ResourcesIds): boolean =>
   snapshot.get(resourceId)?.hasActiveProduction === true;
 
+type AutomationEntityType = RealmAutomationConfig["entityType"];
+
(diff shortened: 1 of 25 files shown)


client/apps/game/src/ui/features/infrastructure/automation/model/automation-processor.ts
```diff
@@ -165,14 +268,9 @@ export const buildRealmProductionPlan = ({
   const configuredCustomIds = Object.keys(realmConfig.customPercentages ?? {}).map(
     (key) => Number(key) as ResourcesIds,
   );
-  const resourceIdsToProcess = new Set<ResourcesIds>(producedResourceIds);
-
-  // Reliability fallback: if snapshot misses active production, keep configured custom resources in scope.
-  if (resourceIdsToProcess.size === 0) {
-    configuredCustomIds.forEach((resourceId) => resourceIdsToProcess.add(resourceId));
-  }
+  const resourceIdsToEvaluate = buildResourceIdsToEvaluate(producedResourceIds, configuredCustomIds);
```
ponderingdemocritus, line 353: This now lets stale `customPercentages` affect Smart planning. `setRealmPreset` preserves `customPercentages`, so a realm that was custom with, for example, `KnightT2` configured and then switched back to Smart still includes that inactive custom key here. `calculatePresetAllocations(resourceIdsToEvaluate, "smart")` then treats T2 army as present, which changes the Smart weights for active T1 troops and can produce fewer cycles plus irrelevant inactive-resource skips in Smart mode. Custom-only ids should only be added when resolving a custom run, or Smart allocations should be computed from active production ids only.

client/apps/game/src/ui/features/infrastructure/automation/production-automation-dashboard.tsx
```diff
@@ -46,6 +49,9 @@ const getStatusDotBg = (statusStr?: string): string => {
   }
 };
 
+const getVisibleSkipMessages = (realm: RealmAutomationConfig): string[] =>
+  (realm.lastExecution?.skipped ?? []).slice(0, 2).map(buildAutomationSkipMessage);
```
ponderingdemocritus: This reads skip details from `lastExecution`, but the render condition is based on the current `lastStatus`. The idle path records `status: "skipped"` with `message: "Idle preset"` without updating or clearing `lastExecution`, and `setRealmPreset` preserves prior execution data. After a previous no-executable skip, switching the realm to Idle will show `Skipped: Idle preset` plus stale recipe/inactive-resource reasons from the old execution. Use the current status message, clear/record execution for idle, or otherwise tie these details to the matching status attempt.

client/apps/game/src/ui/features/infrastructure/automation/production-automation-dashboard.tsx
```diff
@@ -200,6 +207,12 @@ const ProductionAutomationContent = ({ compact = false }: ProductionAutomationCo
                     </span>
                   </div>
                 )}
+
+                {realm.lastStatus?.status === "skipped" && skipMessages.length > 0 && (
```
ponderingdemocritus, line 208: This is a user-visible dashboard UX change, but the diff does not update `client/apps/game/src/ui/features/world/latest-features.ts`. The scoped rule in [`client/apps/game/AGENTS.md`](https://github.com/BibliothecaDAO/eternum/blob/60eb88da31d5492205b2949d57304ffcf1099e56/client/apps/game/AGENTS.md#L5-L10) says every client UX change or new feature must add an entry to that latest-features list. Add a top entry for the automation skip diagnostics.

ponderingdemocritus (commented):
Review found three high-confidence issues in the workspace diff.

PR #4831, opened 2026-05-27
Add hyperstructure auto-garrison

ponderingdemocritus:
Adds hyperstructure auto-garrisoning so capturing armies preserve relic cargo and move surviving troops into an available guard slot, using an atomic multicall for unguarded captures and a post-confirmation follow-up when a final guard is defeated.

Adds shared provider/system-call/client transaction support for `attack_explorer_vs_guard_and_garrison` plus focused tests for the call payloads and planning helpers.

Updates same-owner explorer-to-structure transfer eligibility and records the UX change in latest features.

Validation: `pnpm run format`; `pnpm run knip`; focused Vitest coverage for hyperstructure auto-garrison, transfer eligibility, provider attack-and-garrison, and client transaction mapping; `client/apps/game` typecheck.


djizus:
Handled client side in PR #4840

PR #4849, opened 2026-06-04
[codex] Fix worldmap army removal recovery

ponderingdemocritus:
This PR gates explorer troop recovery by comparing the live troop position with the original removal position, preserves deferred removal context through chunk-switch retries, and removes the stale live-update timestamp path. It adds tests for tile versus zero-count recovery, deferred retry context, and the explorer troop update ordering. Root cause: deferred chunk removals dropped removal metadata, so a live troop update could not cancel the deferred deletion once pending metadata was cleared. Validated with focused worldmap tests, the Three.js type check, format, and knip.

### DIFF
--- a/client/apps/game/src/three/scenes/__tests__/worldmap-update-helpers.test.ts
+++ b/client/apps/game/src/three/scenes/__tests__/worldmap-update-helpers.test.ts
@@ -81,4 +81,26 @@ describe("processExplorerTroopsUpdate", () => {
 
     expect(callOrder).toEqual(["hexes", "army", "live-signal", "pending-clear", "recovery"]);
   });
+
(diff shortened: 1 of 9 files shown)


djizus:
Closing as obsolete — the changes in this PR are already present in `next`. The worldmap army-removal recovery work (`shouldRecoverPendingArmyRemovalFromExplorerTroops`, the `armyLastLiveUpdateAt` → `resolveLastArmyTileSyncAt` consolidation, deferred-removal position requeue, and the accompanying tests) all landed via the `next` branch (merge-base #4848). Verified the helper files are byte-identical and every added/removed symbol matches; `next` is a superset with additional later work on top. Nothing here is missing from `next`.

PR #4853, opened 2026-06-10
Claude/great aryabhata b32e0e

ponderingdemocritus:


djizus (approved):
Reviewed the full diff (phases 1, 2 and 5.1 of the graphics optimization plan) plus follow-up fixes:

- Per-chunk terrain generations: all old bump()/current() call sites migrated; clearCache() resets generations alongside the matrices it flushes.
- Dispose hooks for dropped prepared terrain: dispose/cache/apply are mutually exclusive on every path (warp-travel, prewarm, refresh-commit, visual page build), and the attribute pool guards double-release.
- Shared biome GLTFs: render-order stamp makes shared-material mutation idempotent; scenes are only destroyed at full teardown, immediately followed by clearBiomeGltfCache().
- Hardened two findings: owner-less structure count changes now fall back to invalidating via the update's own hex, and animated biomes clone the gltf object graph so each scene's AnimationMixer drives its own morphTargetInfluences (geometry/materials stay shared).

Local validation: 3390/3390 tests pass, prettier clean, production build succeeds. CI green on c879641f87.

PR #4855, opened 2026-06-10
perf(game): trim three.js hot-path allocations that scale with entity count (graphics plan phase 3)

ponderingdemocritus:
## Summary

Memory and hot-path performance work for the Three.js game client (`client/apps/game/src/three`), aimed at low-spec hardware. The changes come out of a deep multi-agent review of the graphics layer (328 files / ~72k lines); the full prioritized plan lives in [`GRAPHICS_OPTIMIZATION_PLAN.md`](client/apps/game/src/three/GRAPHICS_OPTIMIZATION_PLAN.md) (added in the Phase 1 commit).

This PR bundles four phases of that plan. Commits are scoped one-per-phase, so it reviews cleanly commit-by-commit.

## What's in here (by commit)

**`9582acc` — Phase 1: scope worldmap terrain cache invalidation**
Stop the cache-invalidation storms that defeat the terrain cache during exploration/panning.
- Structure count changes now invalidate only the affected chunks instead of flushing the whole terrain matrix cache + global pools.
- Army movement no longer invalidates terrain (verified the terrain cache reads only biomes + structures, never armies).
- Per-chunk generation counters replace a single global counter that marked every cached chunk stale on any tile change.
- The terrain fingerprint is now a compact FNV-1a digest instead of a ~60KB sorted-joined string rebuilt on every cache hit.

**`b667108` — Phase 2.1–2.3: GPU leaks + morph-texture crash**
- Arrival-ghost ring geometries are shared instead of leaked two-per-ghost.
- Prepared-terrain pooled attributes are released on every dropped chunk transition (prewarm / rollback / stale-drop) instead of leaked.
- Morph textures are resized on instance-capacity growth — previously `setMorphAt` indexed past a fixed `Float32Array` once >64 animated army slots existed and threw a `RangeError` on the frame path.

**`49422835` — Phase 2.5: manager destroy hygiene**
`ChestManager`/`ArmyModel`/`InstancedModel` dispose paths now free instanced GPU buffers + morph textures (via `InstancedMesh.dispose()`, which leaves shared geometry/materials intact), with an `isDestroyed` guard on the async chest load; removed a dead `Group` hexception leaked on every grid rebuild.

**`586d378` — Phase 5.1: share biome GLTFs across scenes**
The ~23 biome GLBs were parsed and GPU-resident twice (worldmap + hexception). They're now parsed once via a module cache and shared; only per-scene instance buffers + morph textures differ. Includes an idempotent biome draw-order fix so the shared material resolves the same `renderOrder` in both scenes.

**`26b97886` — Phase 3: hot-path allocation diet (scales with entity count)**
- Per-frame biome resolution (BigInt/simplex noise, twice per moving army) is memoized by hex.
- Attachment-transform sync no longer builds an input object + 5 closures for every visible army each frame (hoisted the membership guard → scales with attachment count, not army count).
- Compact labels resolve the title directly instead of building the full view model (+detailRows) per moving army per frame.
- Structure hover resolves via the existing spatial-bucket index instead of an O(all structures) scan + `Array.from` per pointermove.
- Frame-loop micro-churn: diagnostics skip the per-frame window snapshot when unchanged; the 1 Hz stamina recompute is gated behind the existing armies-tick policy; the visibility manager mutates its cache record in place instead of allocating on every miss.

## How it was built

- TDD throughout: new pure logic (fingerprint digest, per-chunk generation, morph-texture resize, prepared-terrain dispose decisions, biome/structure-hex caches, title resolvers, diagnostics gating) lands with unit tests; GPU-lifecycle and allocation refactors are behaviour-preserving and covered by existing tests + careful reading.
- Behaviour-preserving by design: cache-invalidation changes were verified against what the terrain cache actually depends on; the attachment and stamina changes reuse guards/policies that already existed.

## Deferred (intentionally, to focused follow-ups)

- **1.5** compose-once-per-batch terrain presentation (590-line runtime API refactor; needs runtime visual QA).
- **2.4** world-update subscription contract (cross-package: `packages/core/world-update-listener.ts` must return its unsubscribers).
- **3.5** TileOpt hydration spatial index (needs a new ECS-stream index; a bug there means missing terrain/structures → needs runtime QA).
- 2.5 tail (InteractiveHex/HighlightHex cleanup, shared-material ownership) — documented in the plan.

## Reviewer notes / test status

- **CI must run `vitest` + `tsc`.** These were authored in a sandboxed worktree where node is killed, so the red→green cycle was verified by construction, not executed locally. Nothing else gates correctness of the new tests.
- **Runtime QA recommended before merge:**
  - Phase 1: pan + explore while founding/destroying structures — no stale/blank terrain; army movement no longer rebuilds neighbour chunks.
  - Phase 2: spawn **>64 concurrent animated armies** (no `RangeError`); watch `MatrixPool`/attribute-pool stats during fast pans (dropped chunks return their attributes).
  - Phase 5.1: eyeball hexception + worldmap terrain, especially transparent biomes' draw order.
  - Phase 3: structure hover still resolves; stamina labels update on tick advance; no visibility/animation regressions.

### DIFF
--- a/client/apps/game/src/three/managers/army-manager.ts
+++ b/client/apps/game/src/three/managers/army-manager.ts
@@ -374,11 +374,14 @@ export class ArmyManager {
           currentTick: currentArmiesTick,
           previousTick: this.lastKnownArmiesTick,
         });
+        // Phase 3.6: stamina is derived from the armies tick, so only recompute it
+        // over all armies when the tick actually advanced (honouring the existing
+        // tick-gate policy) instead of every second. Battle timers keep the 1s
+        // cadence — they already early-out on armies without a battle cooldown.
         if (tickRefresh.shouldRecompute) {
           this.lastKnownArmiesTick = tickRefresh.nextTrackedTick;
+          this.recomputeStaminaForAllArmies();
         }
-        // Update stamina and battle timers every second
-        this.recomputeStaminaForAllArmies();
         this.recomputeBattleTimersForAllArmies();
       } catch {
         // Swallow errors to keep the tick loop alive
@@ -1291,6 +1294,10 @@ export class ArmyManager {
 
     this.visibleArmies.forEach((army) => {
       const entityId = this.toNumericId(army.entityId);
+      // Phase 3.2: skip non-attachment armies before building the input object.
+      if (!this.activeArmyAttachmentEntities.has(entityId)) {
+        return;
+      }
       const instanceData = this.armyModel.getInstanceData(entityId);
       syncArmyAttachmentTransformState({
         entityId,
@@ -2874,8 +2881,11 @@ export class ArmyManager {
         this.updateArmyCompactLabel(army, instanceData.position);
       }
 
-      // 2. Update attachment transforms
-      if (hasActiveAttachments) {
+      // 2. Update attachment transforms. Phase 3.2: only attachment-bearing armies
+      // need the sync (the callee early-returns for non-members), so hoist the
+      // membership check here to avoid allocating the input object + delegate
+      // closures for every visible army each frame.
+      if (hasActiveAttachments && this.activeArmyAttachmentEntities.has(numericEntityId)) {
         syncArmyAttachmentTransformState({
           entityId: numericEntityId,
           army,
--- a/client/apps/game/src/three/managers/army-model.ts
+++ b/client/apps/game/src/three/managers/army-model.ts
@@ -22,6 +22,7 @@ import {
 } from "three";
 import { CSS2DObject } from "three/examples/jsm/renderers/CSS2DRenderer.js";
 import { resizeInstancedMorphTexture } from "./morph-texture-resize";
+import { BoundedHexCache } from "../utils/bounded-hex-cache";
 import { env } from "../../../env";
 import {
   ANIMATION_STATE_IDLE,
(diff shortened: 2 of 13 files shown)


djizus:
Audit note (Aug 13): phases 1, 2, 2.5 and 5.1 of the original plan all landed independently (9582acc4e3, b667108d2b, 49422835ee, 586d378137) — this branch is now exactly the one unlanded phase-3 commit, and the diff applies cleanly against the current feat/single-world-blitz tree. Retitled to match. Plan: merge after #4878 lands on next, with one runtime QA pass that stamina labels still update on tick advance (3.6 gates the per-second recompute, and the stamina-sync stack changed this week).

djizus:
Closing as landed and superseded. The PR's only unlanded commit (26b97886b0) was cherry-picked onto feat/single-world-blitz as 3dd99766cc on Aug 13 and shipped with #4878's branch — verified at the time with the scoped three.js suite (157/158, single failure = the known instanced-model flake) and a clean typecheck. Since then the sync overhaul's S3 phase (PRs #4882–#4884) replaced part of it with something stronger: structure-hex-lookup.ts was deleted in the structures slice, superseded by the WorldSpatialProjection's bucketed spatial index; bounded-hex-cache.ts and the army/model hot-path trims live on in the current tree. Nothing from this PR remains unmerged or unaccounted for.

PR #4859, opened 2026-06-11
Add websocket transaction confirmation

ponderingdemocritus:
Adds an internal transaction confirmation strategy for `packages/provider` that can use Starknet websocket transaction-status subscriptions when a `wsUrl` is configured, while preserving polling as the default and auto fallback path.

The provider constructor now accepts either the legacy retry config or an options object with retry and transaction confirmation settings.

The websocket waiter subscribes to transaction status, treats `PRE_CONFIRMED`, `ACCEPTED_ON_L2`, and `ACCEPTED_ON_L1` as success states, fetches the receipt once, and leaves existing revert/background confirmation handling intact.

Validation: `CI=1 pnpm --filter @bibliothecadao/provider test`, `pnpm --filter @bibliothecadao/provider build`, `pnpm run format`, and `pnpm run knip`.

### DIFF
--- a/packages/provider/package.json
+++ b/packages/provider/package.json
@@ -26,10 +26,12 @@
     "@scure/starknet": "^1.1.0",
     "chalk": "^5.4.1",
     "eventemitter3": "^5.0.1",
-    "starknet": "catalog:"
+    "starknet": "catalog:",
+    "ws": "^8.19.0"
   },
   "devDependencies": {
     "@types/node": "^20.11.10",
+    "@types/ws": "^8.18.1",
     "tsup": "^8.0.2",
     "typedoc": "^0.26.11",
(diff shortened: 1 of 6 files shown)


djizus:
Closing after a full audit (Aug 13). The confirmation pipeline this patches was rebuilt: the provider now polls receipts at 200ms resolving at PRE_CONFIRMED (0.4–0.7s end-to-end on the appchain), with structured failure payloads and an error classifier — the index.ts hunks here target deleted code, and transaction-confirmation.ts never landed. The CDK appchain stack also exposes no wss endpoint, and starknet v8 (now pinned) ships WebSocketChannel natively — if WS confirmation ever becomes worthwhile it's a ~50-line fresh implementation, not this hand-rolled client.

PR #4861, opened 2026-06-15
[codex] Prefer configured node RPC URLs

ponderingdemocritus:
Makes VITE_PUBLIC_NODE_URL required for the game client and removes runtime fallback paths that silently selected hard-coded local or slot RPCs. Updates game and mobile providers so their node URL comes from env/configured runtime state, and changes mainnet/sepolia runtime resolution to prefer a compatible VITE_PUBLIC_NODE_URL before Cartridge defaults. Adds resolver coverage for local RPC selection and env-preferred mainnet fallback. Validated with focused vitest, pnpm run format, pnpm run knip, and git diff --check.

### DIFF
--- a/client/apps/eternum-mobile/src/app/dojo/context/dojo-context.tsx
+++ b/client/apps/eternum-mobile/src/app/dojo/context/dojo-context.tsx
@@ -44,7 +44,7 @@ const useRpcProvider = () => {
   return useMemo(
     () =>
       new RpcProvider({
-        nodeUrl: env.VITE_PUBLIC_NODE_URL || "http://localhost:5050",
+        nodeUrl: env.VITE_PUBLIC_NODE_URL,
       }),
     [],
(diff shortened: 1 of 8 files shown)


djizus:
Closing after a full audit (Aug 13). The env/RPC sweep landed the equivalent behavior: Slot chains are excised from env.ts (VITE_PUBLIC_CHAIN is now sepolia/mainnet/local/appchain, no hard-coded Slot fallbacks), the appchain chainKind takes RPC directly from configured URLs, and profile-builder.ts already prefers a compatible env RPC over Cartridge defaults. As written the diff would also delete hasPublicNodeUrl, which the current tree depends on. One genuine leftover — eternum-mobile still hard-codes SLOT_RPC_URL constants — is a small fresh cleanup when mobile gets its sweep, not a rebase of this.

PR #4864, opened 2026-06-22
[codex] Add AWS game runtime deployer

ponderingdemocritus:
This adds an AWS ECS runtime path for game Torii/Katana infrastructure, including Terraform foundation, runtime image assets, and a manual AWS runtime deployer workflow.

The existing factory Torii deployer, game launch workflow, and indexer maintenance flow now route runtime operations through the clean deployer AWS provider while keeping Slot available as an explicit rollback provider.

Run-store artifacts now persist AWS runtime details so agents and operators can inspect what was created, resized, deleted, or refreshed.

Validation: `pnpm run format`, `pnpm run knip`, `git diff --check`, and `bun test config/deployer/clean/tests/aws-runtime.test.ts config/deployer/clean/tests/indexer-provider.test.ts config/deployer/clean/tests/indexer.test.ts config/deployer/clean/tests/launch-runner.test.ts config/deployer/clean/tests/factory-discovery.test.ts`.


djizus (changes requested):
Requesting changes because the AWS foundation currently encodes the infrastructure topology we are moving away from.

My understanding of the intended target is:

- Slot has reached EOL and is no longer a runtime provider or rollback target.
- AWS provides the shared foundation—VPC, ECS cluster, storage, routing, control table, observability, and recovery.
- Each Blitz game dynamically provisions its own game-owned L3/Katana runtime and Torii runtime within that foundation.
- Eternum remains on Starknet mainnet. It may use AWS-hosted Torii, but does not get a per-game L3.
- Any authenticated Cartridge wallet may launch Blitz.
- Only one Blitz game should be playable at a time. This is a product constraint to protect player liquidity and avoid fragmenting the player base, not an AWS or L3 capacity limitation.
- Manual scheduling is not part of this version.
- The admission slot becomes available when the configured game duration expires. The previous L3 may continue settling and shutting down asynchronously without blocking the next Blitz game.
- If provisioning fails, the admission slot is released immediately.

## Current architecture conflicts with that target

The PR documentation currently provisions one shared Katana for `slot`, one for `slottest`, and allows only Torii on mainnet. It also describes Slot as the rollback provider: [AWS runtime platform](https://github.com/BibliothecaDAO/eternum/blob/b6e7200e01f61fbb99070ac44448ef590e5fa24a/deploy/aws/README.md).

Those assumptions are also enforced in code:

- production Katana is rejected;
- Katana must have `lifecycleClass=shared`;
- Katana cannot be owned by a game or launch run.

See [AWS runtime validation](https://github.com/BibliothecaDAO/eternum/blob/b6e7200e01f61fbb99070ac44448ef590e5fa24a/config/deployer/clean/runtime/aws/reconcile.ts#L95).

That prevents the runtime model we actually need for Blitz: an ephemeral, game-owned Katana L3 under the production Blitz environment.

The workflows also default `RUNTIME_PROVIDER` to `slot`, while the new workflow setup assumes AWS credentials and the factory Torii workflow rejects non-AWS providers. Since Slot is already EOL, maintaining a nominal rollback path adds complexity without providing a usable rollback.

I think we should avoid building and operating this shared-Katana intermediate state if we already know per-game Blitz L3s are the destination.

## 1. Model a Blitz deployment as one game stack

The top-level unit should be a game stack, with Katana and Torii as child runtimes—not two unrelated runtime operations.

A game-stack record should carry:

- immutable game-stack ID;
- authenticated Cartridge wallet identity that requested it;
- game name and approved configuration;
- configured start/end timestamps;
- L3 chain identity and RPC endpoint;
- settlement configuration/identity;
- Katana runtime identity;
- Torii runtime identity and endpoint;
- lifecycle state and structured failure information.

A useful lifecycle would distinguish:

`reserving → provisioning-l3 → deploying-world → provisioning-indexer → ready → active → settling → closed`

with an explicit `failed` state.

The existing per-runtime identity, mutation leases, checkpointing, routing, snapshots, and deletion audits remain useful. They should sit underneath the game-stack orchestration.

For production Blitz specifically, the validation should allow:

- game-owned Katana;
- ephemeral Katana lifecycle;
- Katana under the production Blitz environment.

The restriction against Katana for Eternum/mainnet should remain.

## 2. Add authoritative Blitz admission

The current DynamoDB control table coordinates mutations for one runtime name and allocates ALB capacity. It does not coordinate whether another playable Blitz game already exists: [runtime control](https://github.com/BibliothecaDAO/eternum/blob/b6e7200e01f61fbb99070ac44448ef590e5fa24a/config/deployer/clean/runtime/aws/control.ts#L124).

We need a separate authoritative Blitz admission record, updated with a conditional/transactional write before any infrastructure is provisioned.

The admission behavior should be:

- accept a launch only when no non-expired Blitz admission exists;
- record the game-stack ID, requester wallet, start time and `activeUntil`;
- reject concurrent launch attempts deterministically so only one wins;
- release the admission immediately if provisioning fails;
- consider the admission expired at the configured game end;
- allow the next Blitz game once that time is reached, even if the previous L3 is still settling or being cleaned up.

The play-admission lifecycle and infrastructure lifecycle must therefore be separate. It is valid to have:

- game A in `settling`;
- game B in `active`.

Cleanup for game A must use its immutable game-stack/runtime identities so it cannot affect game B.

The GitHub-backed factory run indexes can remain an operator-facing read model, but they should not be the concurrency boundary.

## 3. Authenticate public launch requests

“Anyone can launch” should mean any authenticated Cartridge wallet—not any anonymous HTTP caller.

At present, the factory create endpoints can dispatch launch workflows without admin authorization or wallet authentication: [factory routes](https://github.com/BibliothecaDAO/eternum/blob/b6e7200e01f61fbb99070ac44448ef590e5fa24a/config/deployer/clean/run-store/cloudflare-worker.js#L86).

The server should verify the Cartridge wallet session/signature before attempting admission and record the verified wallet identity with the game stack. Client-side wallet checks alone are not sufficient because the Worker endpoint can be called directly.

The production server must also enforce the allowed Blitz configuration. In particular, it should not trust arbitrary client-provided values for:

- game duration;
- future start time;
- development or sandbox mode;
- other configuration capable of occupying the single playable slot for an excessive period.

The current factory catalog includes long-running development presets, and the Worker accepts client-provided start/duration values. Those may remain available to internal development flows, but an authenticated public launch must use server-approved production bounds.

Manual arbitrary scheduling, series, and rotation admission do not need to be solved in this PR. The first version should support one immediate or server-determined next-start Blitz launch.

## 4. Provision and publish the stack coherently

A successful launch should perform this sequence:

1. authenticate the Cartridge wallet;
2. atomically acquire Blitz admission;
3. create the game-stack record;
4. provision the L3/Katana runtime;
5. deploy and configure the Blitz world;
6. provision Torii against that L3;
7. verify both runtime and indexer readiness;
8. publish the L3 RPC and Torii endpoints together;
9. mark the game ready/active.

The runtime registry should publish only a complete, ready game stack. A partially created Katana or Torii should never become the active client alias.

For an AWS production deployment, the registry cannot remain optional or fall back to decommissioned Slot endpoints. The client currently permits an empty registry URL and retains Slot-based endpoint defaults: [runtime endpoint configuration](https://github.com/BibliothecaDAO/eternum/blob/b6e7200e01f61fbb99070ac44448ef590e5fa24a/client/apps/game/src/config/runtime-endpoints.ts), [environment defaults](https://github.com/BibliothecaDAO/eternum/blob/b6e7200e01f61fbb99070ac44448ef590e5fa24a/client/apps/game/env.ts).

AWS production should fail configuration or readiness checks when the registry is unavailable rather than silently resolving obsolete endpoints.

## 5. End gameplay separately from settlement and teardown

At the configured Blitz end:

- expire the playable-game admission;
- stop exposing the game as the active playable Blitz;
- transition its stack into `settling`;
- allow another Blitz launch immediately;
- continue checkpointing, settlement verification, and resource cleanup asynchronously.

We should not wait for full L3 closure before admitting the next game because settlement may take additional time.

Deletion should occur only after the runtime’s own settlement and snapshot requirements are satisfied. A delayed or failed cleanup should produce an operator-visible state and alert, but it must not restore or extend the expired player-liquidity lock.

## 6. Keep Eternum separate

Eternum remains on Starknet mainnet and should not enter the Blitz L3 lifecycle.

AWS-hosted Torii for Eternum is still reasonable, but launching an Eternum season costs real funds and should be operator-restricted. The current public factory route accepts `mainnet.eternum` without an admin check.

This is lower priority than the Blitz migration, but the PR should at least avoid widening that exposure. A small server-side guard restricting Eternum creation to authorized operators would be sufficient; a larger Eternum lifecycle redesign is not requested here.

## Acceptance tests

At minimum, I would expect tests proving that:

- an unauthenticated launch is rejected before workflow or AWS activity;
- an authenticated Cartridge wallet can acquire an empty Blitz admission;
- two concurrent authenticated launches result in exactly one accepted game;
- another launch is rejected while the current game’s configured window is active;
- a provisioning failure releases the admission;
- a new game is accepted after `activeUntil`;
- a new game can run while the previous L3 is still settling;
- cleanup for an older game cannot delete or mutate the newer game;
- Katana and Torii are published only after the complete game stack is ready;
- production configuration cannot resolve an EOL Slot fallback;
- Eternum cannot provision a Katana L3;
- unauthorised users cannot launch a mainnet Eternum season.

## Suggested merge boundary

The reusable AWS foundation work—networking, ECS reconciliation, immutable runtime identity, routing, snapshots, observability, backups, DR, and registry primitives—is valuable.

However, the current shared-Katana and Slot-rollback assumptions are foundational invariants, not superficial rollout details. Merging them means implementing and maintaining an intermediate topology we already know is obsolete.

I recommend adjusting this PR so the shared AWS foundation remains, but the dynamic runtime model targets per-game Blitz L3 stacks directly. Manual scheduling and Eternum L3 support are explicitly out of scope.

djizus:
Closing after a full audit against the current tree (Aug 13). The appchain that actually shipped is the in-repo CDK stack (deploy/appchain/cdk: foundation/dev/dns stacks + torii-s2), built on the EC2 + instance-attached-EBS sovereign-katana topology — per my Jul 15 CHANGES_REQUESTED, this PR encodes the shared-Katana/Slot-rollback topology we moved away from, and deploy/aws + config/deployer/clean/runtime no longer match the reorganized deployer. Salvage worth keeping in mind before this branch ages out: checkpoint-agent.sh + runtime-snapshot.mjs (the CDK stack has no snapshot/DR equivalent yet) and the DynamoDB mutation-lease pattern — reference material for future backup/DR work.

PR #4865, opened 2026-06-23
Reduce explore transaction gas

ponderingdemocritus:
Adds `explorer_explore_and_extract` so provider explore does VRF plus one combined movement/reward system call instead of VRF plus move plus extract.
Optimizes the hot path with packed tile mutations, lazy VRF config reads, and specialized food/weight updates while preserving reward and RNG behavior through active movement tests.
Updates web/mobile session policies for the new entrypoint.
Validation: `scarb test --no-warnings`, `pnpm run format`, `pnpm run knip`, `pnpm --dir packages/provider test --run`, `pnpm run build:packages`, `pnpm --dir packages/client build`, `pnpm run build`, `pnpm run build:mobile`, and `git diff --check`; root recursive tests still have unrelated workspace failures/timeouts noted locally.

### DIFF
--- a/client/apps/eternum-mobile/src/app/dojo/context/policies.ts
+++ b/client/apps/eternum-mobile/src/app/dojo/context/policies.ts
@@ -545,6 +545,10 @@ export const policies = toSessionPolicies({
           entrypoint: "explorer_move",
         },
         {
+          name: "explorer_explore_and_extract",
+          entrypoint: "explorer_explore_and_extract",
+        },
+        {
           name: "explorer_extract_reward",
           entrypoint: "explorer_extract_reward",
         },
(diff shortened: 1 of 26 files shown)


djizus:
Closing after a full audit (Aug 13). Every file this touches has since been re-signed with game_id scoping (86a0f5311a and follow-ups) — the diff no longer applies anywhere (git apply --check fails across troop_movement, map, resource, provider, and the dojo_slot*.toml files it edits no longer exist). It's a behavior-preserving gas optimization, and on the self-hosted appchain gas is near-free, so the 3-calls-to-2 win no longer justifies maintaining a second explore path across both game modes. The ~500-line active-movement test suite (RNG-seed storage, reward parity, stamina burn) is genuinely valuable and has no current equivalent — worth re-deriving on the game-scoped models as a fresh PR.

PR #4880, opened 2026-08-13
refactor(client): establish S1 game sync runtime

djizus:
## Summary

Establishes S1 of the sync overhaul as a mechanical, zero-behavior-change ownership refactor:

- adds a headless, session-owned `GameSyncRuntime` with generation fencing and explicit global/player writer lifetimes
- derives the existing global, event, bounded-spatial, bootstrap, player, and game-scope model lists from one executable manifest
- moves the player structure writer lifecycle out of React while preserving its owner subscription, reconciliation, debounce, hydration, and reconnect behavior
- records a reproducible S1 memory/row-count baseline and an exhaustive compile-time `SqlApi` fact-ownership audit

The patch also handles reset-vs-boot supersession as a benign abort without clearing a replacement bootstrap, and makes the player hook safely no-op if its effect runs after runtime disposal.

## Why

This is the first mechanical step toward one session-owned sync runtime feeding RECS. It encodes Client State & Sync Guardrail #1 — **one truth, per fact**: current authoritative game facts live in RECS, while immutable history and genuinely derived aggregates may remain SQL read models only when they are not an alternate copy of a RECS fact.

The executable SQL audit makes every new async `SqlApi` method fail compilation until its ownership is classified.

## Non-goals and follow-ups

- The scene-owned bounded spatial stream remains unchanged in S1.
- The known `Structure` ownership gap closes in S2 when the spatial subscription becomes game-wide.
- S2 will commit the current ad hoc Node import/lifecycle smoke as a repeatable headless acceptance script.
- Wonder/hyperstructure spatial aggregates remain explicit S3 decisions.
- Registered-points and leaderboard SQL endpoints remain an S2/S3 decision because landing pages may run without a populated game RECS session.

## Validation

- `pnpm run format`
- `pnpm run knip` (passes; five existing configuration hints)
- client `pnpm typecheck`
- focused client sync/bootstrap/player suites: 27 passed; final supersession/disposal guard rerun: 5 passed
- core sync suite: 12 passed
- torii ownership audit: 3 passed
- core and torii package builds
- independent full client run: 3,344/3,345; the sole documented instanced-model timeout passed on isolated rerun


djizus:
Merge gate record: knip green after the /game-sync knip paths fix (2893b6bb); test-client green on rerun — the first failure was the katana instance resize mid-scene-smoke (503 from the ALB), not code. Review: approved with 3 asks; 2 fixed in ed6f0859 (SupersededGameSyncStartError handled as benign abort end-to-end, hook switched to non-throwing getActiveGameSyncRuntime), headless smoke script carried to S2 as allowed. Fact-level ownership rule (AGENTS.md guardrail #1) enforced via the SqlApi audit + executable manifest. S2 branches from the updated base: feat/sync-s2-gamewide.

PR #4881, opened 2026-08-13
feat(sync): move spatial state to the game-wide runtime

djizus:
## What changed

- Replace camera-bounded live state ownership with one session-scoped game-wide runtime: subscribe first, hydrate a paginated snapshot, reconcile component absence, and replay buffered callbacks in client receive order.
- Route entity writes, component tombstones, full deletions, and ephemeral event effects through a scheduler-backed RECS ingest queue. Structure now shares the same stream as the other game-wide models.
- Keep the full bounded implementation behind one rollback adapter. `VITE_PUBLIC_WORLDMAP_BOUNDED_SPATIAL_SYNC=true` is now an explicit rollback; omitted flags default to S2.
- Add a repeatable Node headless smoke, runtime metrics, the recovery contract, and the live measurement artifact.

## Why

The bounded writers split ownership across bootstrap, player, and scene lifecycles. Camera movement, reconnects, and competing Torii channels could therefore change sync correctness instead of only changing what Three.js renders.

This enforces the fact-ownership rule for current on-chain game state: the session runtime is the writer, RECS is the sole client truth, and Three.js only selects visible RECS entities. Torii SQL projections and out-of-session landing-page reads are unchanged here and remain explicit S3/S4 adjudication work.

Recovery is intentionally convergent, not gap-free: Torii exposes no universal server revision or transactional multi-page snapshot. The contract documents both page/update race directions and the recovery bound.

## Review fixes included

- Event replay identities are `model:hashed_keys:timestamp`, so an exact replay is suppressed while a later battle, reward, or chest event with the same on-chain keys still fires.
- The legacy bounded flag defaults to `false`, preventing arms that omit the variable from silently selecting the rollback path.

## Impact and rollback

The client holds the full fixed game scope in RECS and camera movement performs no subscription update. The live smoke hydrated 2,472 entities across five pages; the recorded maximum scheduled batch was 23 ms (24 ms on the independent reviewer rerun). Set `VITE_PUBLIC_WORLDMAP_BOUNDED_SPATIAL_SYNC=true` to restore the complete legacy path while S2 is gated.

## Verification

- `pnpm run format`
- `pnpm run knip` (exit 0; five existing configuration hints)
- `pnpm --dir packages/core run build`
- `pnpm --dir packages/core exec vitest run src/sync` — 21/21
- `pnpm --dir client/apps/game run typecheck`
- `pnpm --dir client/apps/game exec vitest run src/utils/public-env-startup.source.test.ts` — 2/2
- Independent full client rerun: 3,353/3,354; only the documented instanced-model flake
- Live headless smoke: 2,472 entities, five pages, correct occupancy decode

## Standing S3 gate

Capture a non-zero `peakLiveUpdatesPerSecond` during an active battle playtest before S3 merges. The completed-game smoke correctly recorded zero and does not fabricate this acceptance result.


djizus:
Merge gate record: both must-fix items verified in 9eb78f8e0e — event dedupe identity now includes the payload timestamp (repeat rewards/battles between the same parties fire correctly; loud throw for any future event model missing timestamp), and the flag default flipped to false so every arm env selects S2 and rollback is always explicit. Doc nit addressed (bidirectional replay staleness bound). Independently verified: core sync 21/21, tsc clean, knip 0, full suite 3,354/3,355 (documented instanced-model flake only), headless smoke run live by reviewer (2,472 entities / 5 pages / 24ms max batch). Standing condition carried forward: non-zero peakLiveUpdatesPerSecond must be captured during the next live playtest BEFORE the S3 PR merges. Fact-level ownership rule (AGENTS.md guardrail #1) enforced via manifest + gamewide channel.

PR #4882, opened 2026-08-14
feat(sync): render chests from the world projection

djizus:
## What changed

This is the first S3 projection slice. It adds a session-owned, headless `WorldSpatialProjection` derived from RECS `TileOpt` state and moves chest rendering onto that projection.

The projection supports full rebuilds, change-diffed add/remove/move notifications, and bucketed bounds lookups. `GameSyncRuntime` owns its install/dispose lifecycle, while `ChestManager` now subscribes only to projected chest changes.

The former chest truth/cache paths are deleted: the scene-local `chestHexes` map, both chest-specific Torii subscriptions, the linear hover fallback, chunk chest hydration, and the repair fetch.

## Why

Chest state had multiple owners across RECS, Torii subscriptions, and scene-local caches. Those paths could disagree, which made stale or alt-layer ghost chests possible.

RECS remains the sole owner of current entity truth. The projection is a disposable, fully rebuildable spatial index over RECS; it does not introduce a second authoritative store. The alt-layer eligibility rule lives in the renderable resolver, so invalid chest state cannot enter the renderer.

## User impact

Chest meshes, hover labels, and move/removal updates now converge through one session-scoped path. Camera movement only filters the freshest projected state and no longer triggers chest repair work.

## Verification

- `pnpm run format`
- `pnpm run knip` (exit 0; five existing configuration hints)
- core sync suite: 28/28
- client typecheck: clean
- full client suite: 3,354/3,355; only the documented instanced-model timeout flake
- live headless smoke: 2,472 entities across five pages, expected chest 10,775 decoded and projected at the requested coordinate
- independent reviewer verification and approval

## Scope and rollout notes

Structures follow in a separate S3 PR. Armies remain a later slice and will use incremental per-entity updates unless measurement justifies rebuilds. In explicit legacy rollback mode, deleting the chest repair fetch means chest-staleness healing is reduced; that is an accepted emergency-path tradeoff until S4 removes the legacy adapter.

### DIFF
--- a/client/apps/game/scripts/run-game-sync-headless-smoke.mjs
+++ b/client/apps/game/scripts/run-game-sync-headless-smoke.mjs
@@ -2,12 +2,13 @@
 
 import {
   GameSyncRuntime,
+  WorldSpatialProjection,
   buildGameSyncModelKeysClause,
   createTimerGameSyncScheduler,
   getGameSyncModelsForChannel,
 } from "@bibliothecadao/eternum/game-sync";
 import { tileOptToTile } from "@bibliothecadao/eternum";
-import { defineContractComponents } from "@bibliothecadao/types";
+import { TileOccupier, defineContractComponents } from "@bibliothecadao/types";
 import { createWorld, getComponentEntities, getComponentValue, removeComponent } from "@dojoengine/recs";
 import { createClient } from "@dojoengine/sdk";
 import { setEntities } from "@dojoengine/state";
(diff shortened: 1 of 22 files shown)


djizus:
Merge gate record: commit 697697ed82 is byte-identical to the reviewer-verified tree (approved, no must-fixes). Reviewer-run verification: core sync 28/28, tsc clean, knip 0, client suite 3,354/3,355 (documented instanced-model flake only), live headless smoke with --expect-occupier-id 10775 exit 0 — projection reported 4 chests in game 13, matching the reviewer's independent torii decode. Alt-filter is built into the renderable resolver (ghost-chest class now structurally impossible). Known trade: legacy rollback mode loses the chest repair fetch (emergency path, dies in S4). Next: feat/sync-s3-structures per docs/plans/sync-s3-slices-codex-brief.md (landed in this PR). Fact-level ownership rule (AGENTS.md #1) upheld: projection stores identity+location only; panels read RECS directly.

PR #4902, opened 2026-08-21
fix(client): harden rendering against freeze failures

djizus:
## Incident

P10 introduced several independent ways for rendering to stop converging under late-match update load. The Aug 21 game-28 incident convicted a synchronous structure-slot drain as the immediate freeze, while renderer failures were also silent and permanently stopped the frame loop. Production remains rolled back to `37ada08e7a` pending this PR and an operator deploy.

## Slice status

- [x] P11A — terminate empty structure-slot drains
- [x] P11B — make renderer failures observable and keep one frame loop alive
- [x] P11C — make sustained-update reconciliation converge
- [x] P11D — close stranded transition/model-load fences
- [x] P11E — delete ineffective dirty-range uploads
- [x] P11F — delete ineffective local-view texture prewarm

The original A–F slices remain separate commits. Review corrections are consolidated in `89b00ab9aa` because they span the ownership seams the deep review exposed.

## P11A — empty structure-slot drain

Require a non-empty slot array before trimming trailing empty slots. The regression removes a bucket's last visible instance and asserts termination, `removeInstance(0)`, `setCount(0)`, and deletion of both bookkeeping entries.

Red before the guard:

```text
timeout 20s pnpm test src/three/managers/structure-manager.lifecycle.test.ts
Result: process made no progress and was killed after 20 seconds, exit 124
```

Green with the guard:

```text
Test Files  1 passed (1)
Tests       11 passed (11)
```

P10 loop census: no other introduced compaction/drain loop can remain synchronously true without advancing. The older `waitForWorldmapHydratedRefreshQueueIdle` yields through `setTimeout(0)` and therefore cannot busy-hang the main thread; an indefinite wait remains a separate deadline follow-up.

## P11B — renderer failure diagnostics and loop ownership

- Resolve `renderer.backend.device` only after the real renderer `init()` completes, attach loss/error diagnostics to that device, and publish `deviceStatus: "ready"` through the same initialization handoff.
- Keep exactly one owned rAF chain. `animate()` is idempotent, the pending handle is tracked/cancelled, and fast fallback failure cannot start a second chain.
- Always schedule after a frame attempt. A `false` render does not clear failure history, and an error reporter failure cannot skip `lastTime` advancement.
- Track alternating failure fingerprints independently. Repeats report at 60, 120, 240… frames, capped at one report per 3,600 frames.
- Emit structured console and tagged Sentry events before recovery eligibility is checked; unpause after fallback-init failure.

Still visible but unrecovered:

- Recovery remains one-shot. A second loss is reported but does not rebuild the backend again.
- A loss/recovery during scene preparation does not explicitly restart rendering through the prepared-scenes gate.
- WebGL context loss after successful fallback is not subscribed or recovered.
- If fallback initialization fails, rendering resumes and the failure is reported, but the lost WebGPU device is not restored.

## P11C — sustained-update convergence

- Component/projection updates enqueue only affected entity IDs; true global changes retain explicit full refreshes.
- Record targeted deltas before the uncommitted-authority gate. Every stale/non-committing pass merges its work back; a partially throwing diff upgrades the retry to a full refresh.
- Resolve each caller when the pass containing its requested version commits. Later sustained requests continue in the background instead of holding an earlier critical chunk caller through the full drain.
- The burst regression now enters through `requestVisibleStructuresRefresh`, holds a real preload boundary, and asserts committed instance state while 100 follow-up requests remain pending.
- Free-slot hints are range/occupancy checked in O(1), and AddressName refreshes prune labels whose structures vanished.
- Critical catch-up logs state `converged`, `failed`, or `timed_out` from the actual race result.

## P11D — stranded fences and terrain authority

- Army finalization clears `isArmyChunkTransitioning` on destroyed, losing, and winning exits; only the winner drains deferred queues.
- A hard timeout records the exact `{timedOutTransitionToken, recoveryTransitionToken}` authority pair. Prepared terrain may commit only under its own token or that exact recovery token; an ordinary `+1` switch and every later token reject and dispose it.
- Manager catch-up receives the prepared token, never the current successor token, so it cannot poison `transitionChunkByToken` with a stale chunk.
- Hexception building work waits with `allSettled`, so one rejected model load no longer poisons later reconciliation.

Accepted residuals: terrain prepared before authority advances beyond its recorded recovery is disposed; if the newer transition also fails, convergence relies on the existing hard-timeout/forced-refresh recovery. The losing army finalizer also has a one-frame window before the winning successor claims the flag; the winning path remains the sole deferred-drain owner.

The pre-existing unbounded `gltfLoader.load` callback is deferred. A queue-level timeout would release the lane while an abandoned callback could still mutate models later; the correct follow-up is one shared asset-loader deadline with cancellation or late-result disposal.

## P11E — dirty-range decision: revert

The range helper and its instance-attribute call sites are deleted. Three r185 clears ranges on its internal instance mirror rather than these source attributes, and the WebGPU uniform path for the affected meshes ignores application ranges. There was no measured upload reduction to preserve.

The three `InteractiveHexManager` clear/empty paths retain `instanceMatrix.needsUpdate = true`; tests cover each path. `path-renderer.ts` deliberately keeps `updateRanges` because those are plain `BufferAttribute`s, not r185 instance-mirror attributes.

## P11F — prewarm decision: delete

The measured production working set is about 200 MiB/180 textures, while the replacement 64 MiB policy skipped the operation all-or-nothing. That made the controller, deadline, lifecycle latch, and upload loop unreachable machinery rather than a real bound. Following the P6D precedent, the prewarm module, lifecycle wiring, scene texture-discovery API, and dedicated tests are deleted. With no latch remaining, visibility changes and device-loss recovery cannot re-arm speculative uploads.

## Validation

```text
pnpm test
Test Files  702 passed (702)
Tests       3102 passed (3102)

pnpm test <11 review-focused files>
Test Files  11 passed (11)
Tests       117 passed (117)

pnpm typecheck
pnpm run format
pnpm run knip
git diff --check
```

All pass. Knip reports five existing configuration hints and no violations.

## Deployment gate

Do not deploy this branch piecemeal. P11A–F merge together; production stays on `37ada08e7a` until this PR merges and the operator deploys.


djizus:
## P11 consolidated review — request changes

Four independent deep reviews plus local verification (focused suites re-run green). Slice verdicts: **A, B, E pass their gates; C, D, F fail theirs.** The failures share one theme: the mechanism landed, but an edge the brief convicted (or its mirror) survives. Blockers first, then required fixes, then non-blocking notes. Line refs are branch tip `194e9fc6b3`.

### Blockers (fix before merge)

**BL-1 (P11D, HIGH ×2) — the `+1` token rule is unsound.** `canCommitPreparedTerrain` (`worldmap-owned-terrain-commit.ts:29-33`) accepts `prepared + 1` as "timeout recovery", but an ordinary chunk switch also bumps by exactly 1 (`worldmap.tsx:7075` vs `worldmap-chunk-transition-runtime.ts:48`). Consequences: (a) a queued commit for chunk A paints stale terrain over chunk B and moves `currentChunk` backwards — the exact condition this slice exists to eliminate, now reached by committing instead of dropping; (b) `warp-travel-chunk-switch-commit.ts:106-108` then forwards the *current* token with the *stale* chunk, so `transitionChunkByToken` binds token N+1 → chunk A and the real chunk B's manager catch-up is silently rejected (`manager-update-convergence.ts:56`) — armies/structures/chests missing until the next transition. `worldmap-owned-terrain-commit.test.ts:65-77` asserts the defective contract via a generic `supersede` bump. **Fix:** record the actual recovery token when a hard timeout fires (it already exists at `worldmap.tsx:7033`) and accept only that recorded token; pass the *prepared* token, not the current one, to `scheduleManagerCatchUp`. Rewrite the test so recovery and ordinary supersede are distinguishable.

**BL-2 (P11C, HIGH ×2) — targeted refresh deltas are consumed-and-dropped.** `structure-manager.ts:259-266` drains `pendingVisibleStructureRefreshIds`/`shouldRefreshExistingStructures` into locals before the pass runs; every non-committing exit (`:1172`, `:1185`, `:1321`, task rejection) discards them with no re-queue, and `previousVisibleIds` survives chunk crossings — so a realm that levels up during a chunk crossing renders at the old level indefinitely. The old `refreshExisting: true` default self-healed this; that heal was removed in the same commit. Worse, `:1109-1111` returns on an uncommitted chunk *before* `:1116-1119` records the ids, so every targeted update during stall recovery is permanently lost. **Fix:** record the pending delta before the uncommitted-chunk gate, and merge it back into pending state on every non-committing exit. Add the missing failure-direction test: discarded pass → ids re-queued → eventually committed.

**BL-3 (P11C, MED-HIGH) — the convicted symptom survives.** `createCoalescedAsyncUpdateRunner` (`manager-update-convergence.ts:86-106`) drains unbounded and mid-flight callers await full drain, so sustained RECS rates above ~1/pass still starve `updateChunk` past the 12s timeout → recovery → new transition token — the `transition=1..15` signature. And the burst test (`manager-update-convergence.test.ts:41-70`) is tautological: it never touches `requestVisibleStructuresRefresh`, never invalidates, and would still pass with the old fence behavior. **Fix:** resolve the critical caller once *its* delta has committed (or bound the drain), and drive the convergence test through `requestVisibleStructuresRefresh` under sustained updates, asserting committed instance state.

**BL-4 (P11B, HIGH) — double rAF chain after fallback failure.** `handleDeviceLossFallbackFailure` (`game-renderer.ts:353-362`) unpauses and calls `this.animate()` while the pre-pause rAF callback may still be pending (no handle tracking, no `cancelAnimationFrame` in the file). A fast fallback rejection — the likely kind: cached module promise resolves in a microtask, then constructor/getContext throws — lands before the next frame boundary and both chains survive, doubling frame work on a machine that just lost its GPU; it re-doubles per occurrence. The test stubs `subject.animate` so it cannot catch this. **Fix:** make `animate()` idempotent (an `isAnimationLoopRunning` flag or stored rAF handle) and test restart-while-pending.

**BL-5 (P11F, MED-HIGH) — the 64MiB budget is a silent kill switch, not a bound.** `local-view-texture-prewarm.ts:102` skips all-or-nothing on the estimated total, and the only production measurement (`logs.txt`: `estimated_mb=200.0`, `textures=180`) exceeds it on every session — the whole module (runtime ceiling, deadline, upload loop) is now unreachable dead machinery, while the commit message reads as throttling. **Fix (binary):** make the budget truncating — upload a 64MiB prefix and report the partial — and cite the 200MB measurement in the PR; or invoke the P6D precedent and delete the module. If bounded rather than deleted, also fix: the resident-texture ceiling double-counts the prewarm's own uploads (`:169` vs the entry gate at `:99` — use a delta budget, and check "no more textures → completed" before the ceiling so the final tick can't flip completed→cancelled), and arm the 10s deadline on first deferral rather than at `start()` so `interaction_deadline_exceeded` doesn't misreport.

### Required fixes, non-blocking severity but same round

- **P11B:** breaker resets on any fingerprint change — alternating errors (e.g. varying "Vertex buffer slot N" indices) defeat the throttle entirely (`renderer-animation-runtime.ts:80-90`); add a fingerprint→count map or global budget. Add backoff/cap on the 60-frame repeat report (~3.6k Sentry events/hour/user in the steady failure state). Don't call `onFrameSuccess` when `renderFrame` returned false (`game-renderer.ts:628`) — a no-scene frame resets the breaker. Wrap the frame-error reporter so its own throw can't skip the `lastTime` assignment (`:616`). Add the missing `deviceStatus: "ready"` assertion.
- **P11C:** partial-commit repair — `bindStructureInstance` capacity throw mid-diff leaves removed instances gone and ids `staying` forever (`manager-visibility-diff.ts:53-64`); make the commit path restore or re-queue on throw. Guard the free-slot hint (`:1438-1451`) with an O(1) `slots[freeSlot] === undefined` check so map divergence can't create sparse phantom instances. AddressName updates no longer prune labels for vanished structures (`:998-1001` vs old `:396`).
- **P11D:** `≥ +2` supersede still drops with no convergence rule (narrowed livelock, not removed) — either close it or state the accepted residual in the PR body alongside the one-frame unfenced window between a losing and winning army transition.
- **P11E:** restore the three dropped `needsUpdate` lines in `interactive-hex-manager.ts` clear/empty paths (`:346`, `:367-372`, `:429-434`) for contract symmetry, and add the one-line PR note that `path-renderer.ts:502-506` keeps `updateRanges` deliberately (plain BufferAttributes, not subject to the r185 instance-mirror bug).
- **P11F:** add the behavioral test the latch actually needs — after a device loss, `visibilitychange` → visible must not re-arm.

### Clean passes worth keeping as-is

P11A (verified red→green independently). P11B's device-init inversion, honest fixture, and single-`requestNextFrame`-per-path walk all check out — including the bonus honesty win that a silent three-side WebGL fallback no longer marks `deviceStatus: "ready"`. P11C's fence-ordering inversion, options-based call-site audit, free-slot O(N) fix, and honest catch-up log are correct. P11D's army-flag hoist (with the winning-drain exclusivity intact) and `allSettled` conversion are correct, and the GLTF-timeout deferral rationale is accepted. P11E's revert was verified hunk-by-hunk against current code — the three places a blind revert would have regressed were all handled correctly.

Production stays rolled back; the redeploy gate is this PR post-fixes.

PR #4919, opened 2026-09-08
fix(herald): treat a delete for a row the fold never held as a no-op

djizus:
Herald on the lab box has been crash-looping since 20:24 UTC today. The fold throws `delete for ResourceArrival:0x3895c4… has no preceding StoreSetRecord`, and every systemd restart replays from the 20:22 checkpoint (block 344156) straight into the same event. Result: `herald.realms.party` returns 502 on every route and clients lose their game socket.

The chain is allowed to emit that event. `ResourceArrivalImpl::delete` (`contracts/l3/game/src/systems/utils/resource.cairo:729`) calls `erase_model` unconditionally when an arrival day settles to zero, and Dojo emits `StoreDelRecord` whether or not the row was ever written. The fold's delete branch already tolerated a missing row; it was the game-id lookup for the outgoing diff that demanded a prior `StoreSetRecord` and threw.

A delete for a row neither the fold nor its parent overlay holds now returns no change. Update and member events without a prior row still throw, since they cannot be applied without the row's key fields.

Verification: new fold test covers the base fold and an overlay; `apps/herald` suite and typecheck pass. `pnpm run knip` reports pre-existing unused enum members in `packages/types`, none in Herald.

Recovery on the box: `world-fold.ts` is identical between the box's checkout (`client-scale-96p` @ `37b2c1e6c70`) and `next`, so this single commit cherry-picks cleanly there, followed by `sudo systemctl restart herald`.

https://claude.ai/code/session_01RcynR9YYAEgM9y7vDKq1i2

### DIFF
--- a/apps/herald/src/world-fold.test.ts
+++ b/apps/herald/src/world-fold.test.ts
@@ -179,6 +179,25 @@ describe("WorldFold", () => {
     expect(fold.snapshot(7, 12).models[0].rows).toEqual([]);
   });
 
+  it("treats a delete for a row it never held as a chain-legal no-op", () => {
+    const fold = new WorldFold(registry);
+    fold.apply(
+      decodeRequired(
+        registry,
+        rawEvent(WORLD_EVENT_SELECTORS.set, "0x101", ["0x2", "0x7", "0x2", "0x5", "0x3", "0x1", "0x4", "0x5", "0x2"]),
+      ),
+    );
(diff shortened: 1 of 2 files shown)


djizus:
Note on verification: `apps/herald` full suite shows one failure, `model-registry.test.ts > builds a decoder for every executable sync model from the Madara manifest`. It fails identically on plain `next` without this change, so it is pre-existing and unrelated.

PR #4925, opened 2026-09-10
feat(hud): phone layout for the in-game HUD, plus the mobile client scope brief

ponderingdemocritus:
## Why

The client now boots on an iPhone, but the desktop HUD covered the map: two 320px columns, a 300px minimap, a header row wider than the screen, and 1320px workspace panels. This PR gives the in-game HUD a phone layout below Tailwind's `lg` breakpoint and leaves desktop untouched. It also lands the mobile scoping brief that motivated it.

## What changes

**One lane signal.** `apps/game/src/hooks/helpers/use-compact-hud.ts` resolves `null | "portrait" | "landscape"` from two media queries (`(max-width: 1023px)` and the same with `orientation: landscape`). The HUD shell and popover placement both read it; there is no second breakpoint definition anywhere.

**Compact HUD shell.** Below `lg`, `CompactHud` replaces `LeftCommandSidebar` and `BottomRightPanel`:
- Portrait: a bottom tab bar (Empire or Standings for spectators, Map, Log with unread badge, Chat with unread badge, and Details when a tile is selected) with one sheet above it, capped at 55dvh. Details opens on a new selection, the same rule as the desktop panel.
- Landscape: the bar becomes a vertical rail on the right and the sheet a full-height drawer beside it, at most half the width, so the map keeps the left half at full height.
- One thing open at a time; tapping the active tab closes it. The chat window stays mounted so the realtime client keeps its connection, but its strip only shows while open.

The Build / Logistics / Military surfaces moved out of the sidebar into `LeftViewSurfaces` so they render on both layouts. `SpectatorStandings` and `BottomRightPanel` export frameless pieces the sheet reuses; `QuickFeed` exports the feed hooks and notices both layouts share.

**Surfaces dock on phones.** `PopoverPanel` places every anchor as a bottom sheet in portrait and a right-hand drawer under the header in landscape. The seven inline `w-[1320px] h-[calc(100vh-7rem)]` copies (Market, Production, Construction, Military, Logistics, Battle Lab, Chest) became one `SURFACE_WORKSPACE_CLASS` with the portrait and landscape sizes.

**Touch and safe areas.** Tooltips render nothing on coarse pointers (`isCoarsePointer()` in `utils/pointer.ts`, also replacing the duplicate query in the procedural character runtime). The two conflicting viewport metas in `index.html` became one with `viewport-fit=cover`; header, tab bar, rail, sheets and drawers pad for all four safe-area insets.

**Docs.** `docs/plans/mobile-client-scope.md`: the two prior mobile attempts and why they died, what made the renderer crash phones, what ports for free (Herald is JSON over WebSocket, gameplay transactions sign with a local key), and the phased plan this PR starts.

## Verification

- `pnpm typecheck` in `apps/game`: 0 errors.
- `pnpm test` on the hook, popover, tooltip, feed and world containers: 21 files, 85 tests.
- `pnpm -w run knip` and prettier on every changed file: clean.
- Checked on an iPhone in portrait by hand. Landscape geometry (which side the notch lands on, rail tab heights) has only jsdom class and style assertions so far.

## Not in this PR

- Touch input in the three.js scene. Right-click actions and Escape have no touch equivalent yet, so army orders on the world map are still not reachable by touch. Next pass, in `three/managers/input-manager.ts`.
- The internal layouts of the workspace panels. They fit the screen and scroll, but their contents are still desktop grids.
- The renderer mobile lane (force-webgl, DPR 1.0, model compression) described in the brief.

### DIFF
--- a/apps/game/index.html
+++ b/apps/game/index.html
@@ -21,9 +21,13 @@
     <!-- covers -->
 
     <!-- blitz covers -->
-    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no, viewport-fit=cover"
+    />
+    <meta name="apple-mobile-web-app-capable" content="yes" />
+    <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent" />
     <title>Realms: Blitz</title>
-    <meta name="viewport" content="width=device-width,initial-scale=1" />
     <meta name="description" content="The ultimate strategy game on Starknet" />
     <meta property="og:title" content="Realms: Blitz" />
     <meta property="og:description" content="The ultimate strategy game on Starknet" />
--- a/apps/game/src/hooks/helpers/use-compact-hud.test.tsx
+++ b/apps/game/src/hooks/helpers/use-compact-hud.test.tsx
@@ -1,0 +1,56 @@
+import { act } from "react";
+import { createRoot } from "react-dom/client";
+import { afterEach, expect, it, vi } from "vitest";
+import { COMPACT_HUD_MEDIA_QUERY, useCompactHud } from "./use-compact-hud";
+
+const Probe = () => <output>{String(useCompactHud())}</output>;
+
+const render = async () => {
+  (globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
+  const container = document.createElement("div");
+  const root = createRoot(container);
+  await act(async () => root.render(<Probe />));
+  return { read: () => container.querySelector("output")?.textContent, unmount: () => act(async () => root.unmount()) };
+};
+
+const installMatchMedia = (matches: boolean) => {
+  const listeners = new Set<() => void>();
+  const query = {
+    matches,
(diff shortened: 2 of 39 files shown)


ponderingdemocritus:
Pushed `8743ffe6d04`: touch input for the map scene, which closes the first "not in this PR" item.

- One pure `TouchGestureRecognizer` (tap, double-tap, long-press, pinch) under `InputManager`; touch dispatches through the existing listener types, mouse is unchanged.
- Orders on touch: first tap on a reachable hex previews the path and cost, second tap on the same hex (or a long press) commits. Same hover and right-click handlers, one pure policy deciding which a tap means.
- Pinch zoom on the world map feeds the wheel's zoom coordinator with matching sensitivity.
- Canvas drops the iOS callout and text selection on long press.

Gates: vitest on recognizer, policy, InputManager, pinch mapping and HUD containers (144 tests), typecheck, three-types, knip, prettier. Still to check on a real device: click suppression after `preventDefault` on pointerdown in iOS Safari, and two-finger pan coexisting with pinch.

Found while here, unchanged: the owned-structure context-menu store slice has no reader, so that right-click path is already dead. `test:three:changed-guard` never matches from `apps/game` because it compares root-relative paths against a `src/three/` prefix.

PR #4930, opened 2026-09-10
ci: deploy the client on every push to next that touches it

djizus:
The client deploy was manual-only, so `next` ran ahead of play.realms.party until someone dispatched it. This adds a push trigger on `next` filtered to the client's inputs (`apps/game`, `packages`, `config`, the lockfile, the workflow itself), matching the box services' rule. Dispatch stays for a redeploy or a custom version label; on push the label falls back to the SHA as before. The existing concurrency group keeps runs sequential.


djizus:
Superseded by #4929.

PR #4951, opened 2026-09-11
fix(contracts): enforce settlement identity and checkpoint finalization

djizus:
Paid Blitz entry could be reused through a second account, open Eternum could reuse a canonical realm, and ranking could settle every completed hyperstructure inline. Share one owner entitlement across both formats, allocate canonical realms without replacement, and require contiguous checkpoints of at most sixteen structures before ranking.

Includes the audited accounting fixes: settle before multiplier/share changes, indexed completion with idempotent backfill, automatic close-time points, equal-floor tied rewards, per-game faith funding, owner-store faith settlement, and correct generated-point stories. Eternum uses Blitz placement independently of Order; paid metadata is preserved.

Evidence: the three added settlement/ranking regressions fail against 9c07242713b for the intended reasons and pass with this change. Validation: scarb test passes all 210 tests, and scarb fmt passes. No live gas measurement is claimed.

Scope: no balance changes or L2 deployment. Reserved serialized grace and legacy settlement capacity slots remain. The new Eternum entrypoint and checkpointed ranking require matching client and operator support, plus a completed-hyperstructure backfill. The Blitz calldata correction already matches the deployed class. Existing games must be drained; reopening their entry requires a separate settlement backfill.

The independent review verified all 210 contract tests on the complete stack. After renaming the three regression tests, `scarb test systems::registrar::tests` passes all 35 registrar tests (175 filtered out). `scarb fmt` also passes. The final edits change no contract behavior.

## Appchain upgrade procedure

The Blitz calldata correction matches the class already deployed on the box. The new Eternum `settle(game_id, name)` entrypoint and checkpoint requirement for ranking need a coordinated world, client, and operator release. This cutover has not been run.

### Inventory and cutover

1. Record the box's chain ID, world address, class hashes, manifest, preset IDs, and running client/service revisions.
   Read the actual deployed Blitz ABI. Inventory the deployed identity registry, tokens, ledger, and other L2 contracts
   separately; the new L2 endpoints being undeployed does not make existing assets disposable.
2. Keep existing deployed addresses and state. Back up the chain volume, service databases, manifests, and release
   revisions before migration. Do not reset the chain or recreate token contracts as part of this release.
3. Stop new launches and drain existing games before the upgrade. Do not resume entry into a game created before this
   release: its settlements do not populate the new owner entitlement and canonical-realm allocation models. Supporting
   entry into those games requires a separate settlement backfill, with duplicate owners/realm IDs reconciled first.
   Close the daily games still in Registration before upgrading; do not accept new entries into them afterward. Historical games remain available for review.
4. Upgrade the world and its permissions, copy the resulting ABIs, and restart Herald with the new model schemas.
   Existing completed hyperstructures require the completion-index backfill before any new season-close or ranking
   operation. Keep finalization stopped until the migration has verified every affected game.
5. Register immutable Eternum preset 10. Preserve existing Blitz presets 8 and 9 and their balances. Check the factory's
   format-to-preset guard against those registered rows.
6. Deploy the matching launch service, Herald, and client together, then open new games only after the smoke tests below
   pass. Deploy new L2 endpoints only after their ABI and existing dependencies have been reconciled with the inventory;
   this game release does not redeploy existing L2 contracts.

### Backfill and finalization

The chain admin calls the registrar's `backfill_completed_hyperstructures`, `checkpoint_share_points`, and
`rank_players` methods. The prize contract accepts only the registrar contract, not an additional signing key. One
chain-admin account can perform all three steps; the ledger operator remains a separate responsibility.

The box check identified four dev-off daily games with completed hyperstructures that need the backfill before ranking.

Run `bun config/deployer/clean/cli/backfill-hyperstructures.ts --game-id <id> --herald-url <url> --dry-run` first, then run
the same command without `--dry-run` using the chain admin's `DOJO_ACCOUNT_ADDRESS`, `DOJO_PRIVATE_KEY`, and `RPC_URL`.
Archive the confirmed source block, planned IDs, transaction hashes, and final indexed count per game.

After the game ends, checkpoint completed hyperstructures in contiguous batches of at most 16. Retrying an already
checkpointed range is allowed. Ranking checks the completed prefix and end time in constant storage work; it no longer
settles the entire hyperstructure list inline. The batch size is a protocol limit, not a measured gas guarantee. Measure
the worst supported shareholder count and the final roster/result emission before release.

### Compatibility and retirement

`GameRegistry.registration_grace_seconds` remains a reserved serialized slot, initialized to zero. Its client component
slot remains in the same position. It no longer gates scoring or ranking. The legacy settlement capacity fields also
remain reserved to preserve existing serialized presets; the obsolete expansion method is removed.

Retire the legacy per-world deployer as a separate review item. The previous CLI accepted only `madara` and `appchain`;
retirement removes that old configuration replay path. It does not provide a new deployment path for appchain, sepolia,
or mainnet. The supported launch path in this release is the shared Madara world and registrar. Do not present this
retirement as an accounting fix.

### Release gates

- Fresh paid Blitz: two accounts bound to one owner cannot both settle; independent owners can enter.
- Fresh open Eternum: canonical IDs are unique per game, paid reservations share the same allocation pool, metadata
  matches the canonical data, positions remain independent of Order, and entry works after the main start.
- A copied ended game with completed hyperstructures backfills without changing its counters or points; retries do not
  duplicate entries. Checkpoint batches register each interval once; ranking refuses incomplete checkpoints.
- Equal players receive equal whole chests regardless of input order; indivisible leftovers stay unallocated.
- Eternum closes on the real victory condition with dev mode off. A workload that never reaches that condition is not a
  successful lifecycle test.
- Record actual gas/receipt evidence for checkpoint batches, season close, and final roster emission. Verify factory
  creation, settlement confirmation through Herald, and entry into the playable map with the deployed client.

If a gate fails after the world upgrade, keep entry and finalization stopped. Do not roll the client alone back to an
incompatible ABI. Restore the saved stack and state only as one coordinated recovery, or ship a verified forward fix.

Stack 1/5. Depends on https://github.com/BibliothecaDAO/eternum/pull/4950. Review this diff against its base branch; merge and retarget in order after the existing CI PR.


djizus:
Landed on next through #4955 (merge commit 38628a45dc9), which carried the whole stack.

PR #4952, opened 2026-09-11
feat(config): register and launch the Eternum preset on Madara

djizus:
Resolve madara.eternum through the shared environment and immutable preset pipeline, with preset 10 and dev mode off. Remove registration-grace behavior from payload construction and generated balance configuration without shifting the existing onchain storage slot.

Add a completion-index migration planner and CLI that preserves assigned slots, rejects incomplete or cross-game snapshots, and supports dry runs and resumable admin batches. The same chain-admin account forwards backfill, checkpointing, and ranking through the registrar.

Validation: migration planner tests, CLI dry-run and wrong-game rejection, configuration source tests, registrar calldata snapshots, formatting, and the final stack's Knip check. The broader config typecheck still has existing dependency and type errors; this is not reported as a green repository-wide typecheck.

Scope: depends on the contracts change. No preset was registered on a live chain and no existing balances or external token addresses are changed. The legacy deployer retirement is a separate review item in this stack.

Stack 2/5. Depends on https://github.com/BibliothecaDAO/eternum/pull/4951. Review this diff against its base branch; merge and retarget in order after the existing CI PR.

The upgrade procedure and box release gates are in https://github.com/BibliothecaDAO/eternum/pull/4951.


djizus:
Landed on next through #4955 (merge commit 38628a45dc9), which carried the whole stack.

PR #4963, opened 2026-09-12
feat(agents): M0 spikes, M1 shared game-client, M2 agent runner

ponderingdemocritus:
The browser, headless client and lab harness now use the same game client, views and actions. The runner uses that client for settlement, observation and gameplay. The two hired-agent briefs define the remaining control-plane and hosting work; this PR does not implement the native world.

The shared client owns game scope, deployment lookup, Herald ingestion, RECS projection and action submission. The browser keeps rendering and UI state. Explicit world selections must resolve to that deployment; entry, registration and history readers no longer fall back to another world. The RECS bridge uses the shared structure view.

The runner provisions from the Labor building count, handles both restart cases and plans explorers on their own layer. Harness receipt polling and Herald confirmation share one deadline, missing binding authority fails before game creation, and the Blitz script reads directory JSON correctly. Harness imports use a declared workspace dependency. Separate CI jobs run the runner, harness and sandbox-proxy tests; the existing headless fixture gate remains.

The infrastructure documentation now describes the current deployer, RPC and identity service, including `REALMS_SERVER_PORT=3000`. Obsolete Sozo instructions are removed. These corrections supersede #4992. The deploy script is unchanged; seeded-world manifests, local service configuration and preparation evidence remain uncommitted.

Validation at `045147fc6eb`, after merging `next` (`457d8636a3e`):

- Client: 724 files / 3,580 tests; typecheck passed.
- Shared client: 52 targeted tests; runner: 64 tests and typecheck; harness: 45 tests; sandbox proxy: 5 tests.
- Shared-package build and headless fixture passed; the fixture reaches `running` with 34 RECS entities.
- Live laptop smoke: fresh game, one bot, three explores, zero failures or reverts.
- `pnpm run format` and `pnpm run knip` passed. All 18 GitHub checks are green, including the new runtime jobs.

The 96-bot acceptance run, full live Blitz agent game and external sandbox/LLM gates remain unverified. The earlier laptop smoke runs and offline tests do not clear those gates. Merge this shared-client prerequisite before starting the native slice; #4987 remains draft reference only.


djizus (changes requested):
I merged `next` into this branch (34e2f76) so it is current: the moved sync modules now carry the PWA foundation's story-event scope (Herald session takes `worldAddress`, the observer receives `(event, scope)`), the quest views/hook/components went with the quest retirement, the settlement builder gained `devRealmId`, and `selection.profile.test.ts` installs the world directory it mocks. Typecheck, core (265), client (3,455), harness (29), knip and prettier are green on the merged tree.

The M1a moves are mechanical and the entity-key single-instance claim holds in dist. `createGameClient` is a real seam (boots against a fake socket without React/three) and the views are query builders plus pure readers, so one-truth is respected. Requesting changes for the items below.

**Scope**
- The body describes M0 + M1a and says M1b "follows", but e2b04fa (`createGameClient`) and 951f6c1 (views) are on the branch, and the verification line predates them. Update the body, or split those two commits out. Both are complete and green on their own.

**Blocking**
- `packages/core/src/client/game-client.ts`: `GameClient.signer` and `connect()` have no reader (the comment says nothing reads them until actions land); `world`, `presetId`, `projection` are never read by the app either. Land them with the M4 signing lane, not here.
- `apps/game/src/init/bootstrap.tsx`: `getWorldById(profile.worldId) ?? getDefaultWorld()` is a silent default. `worldId` is optional on the profile, so a stored profile without one boots the default world's namespace and Herald URL while `dojoConfig` still uses that profile's own `worldAddress`/`rpcUrl`. The old path could not diverge. Throw on an unresolvable world id; the same fallback appears in eight other app sites, so treat it as a class.
- `docs/plans/hired-agents-milestones.md` (M0 item) points at a continuity ledger under a gitignored `thoughts/` path. That names tooling that must not appear in the repo, and the gate it describes cannot be verified from the repo. Same for the footer on this PR body.

**Should fix**
- `apps/game/src/sync/recs-store-bridge.ts` `readPlayerStructures` re-derives "the player's structures grouped by category" that `views/structures.ts` now defines. Migrate the bridge onto the view.
- The world-directory install-by-import invariant is unstated: any module importing `resolveGameId`/`resolveWorldIdForGame` from the subpath without the app shim throws at runtime, not typecheck. It bit the merge above. Install explicitly at the app entry, or make the shim the only exporter of those readers.
- `packages/core/src/client/world-directory.ts`: boot errors used to name the env var to fix (`VITE_PUBLIC_HERALD_URL` etc.); they now say `Herald URL for world "blitz"`. Pass the host's label through `WorldDeploymentInput`.
- `packages/core/src/client/herald-session.ts`: one input carries an observer with eight hooks plus four loose callbacks of the same kind (`onLiveUpdate`, `onMetrics`, `onSnapshotProgress`, `onSubscriptionActive`). Fold them before more consumers appear.
- With a fresh runtime per client, a superseded bootstrap now rejects with "transport was disposed" instead of `SupersededGameSyncStartError`; `bootstrap.tsx` captures that to Sentry as an unhandled error. Map it, or filter it in bootstrap.
- `apps/agent-runner` and `deploy/madara-lab/scripts/sandbox-proxy-handler.ts` have offline tests (sub-second) but no workflow runs them. Wire them or state the exception in the PR.
- `packages/core/scripts/run-game-client-headless.mjs` offers `--chain appchain`; that deployment is dead.
- The architecture brief anchors on RECS and `setup()` and does not mention the Dojo tooling exit, the native-Cairo fresh-world spike, or RECS leaving the client. Say the store behind `game-client` is expected to change, or the views become one more layer to migrate.

apps/agent-runner/src/entry.ts
```diff
@@ -0,0 +1,188 @@
...
+    usernameFelt: shortString.encodeShortString(username),
+    gameId: game.client.gameId,
+    // The lab world has no VRF provider; settle draws its position without a request_random call.
+    vrfProviderAddress: null,
+    grantStartingTroops: true,
+  });
+
+/** Blitz realms produce nothing until provisioned; a funded T1 balance is the RECS sign it already happened. */
+const ensureProvisioned = async (game: RunnerGame, signer: AccountInterface, structures: ID[]): Promise<void> => {
+  const unprovisioned = structures.filter((structureId) => startingTroopType(game.client, structureId) === undefined);
```
djizus: **P1 — Read provisioning state instead of the troop balance.** Settlement in this same path passes `grantStartingTroops: true`, so a newly settled realm already has T1 troops before its economy is provisioned. This filter therefore skips `provision_realm`, which creates the Labor building, seeds production, and grants starting non-troop resources. The contract's provisioning guard checks the Labor building count (`contracts/l3/game/src/systems/utils/realm.cairo:87`), not the troop balance. After troops are spent, the inverse happens: an already provisioned realm is selected again and the transaction reverts.

Reproduced by calling `ensureSettled` with a settled realm, funded starting troops and no Labor building: it completed with zero provisioning submissions. Use the authoritative provisioning fact and cover fresh settlement plus restart after troop expenditure.

deploy/madara-lab/harness/driver.ts
```diff
@@ -1041,15 +887,27 @@ export async function trackTransaction(options: TrackTransactionOptions): Promis
...
+  // The receipt lifecycle is the measurement (pre-confirmed and L2 timings at the poll boundary); the client's own
+  // confirmation rides alongside so a bot never plans its next step before the client processed this one.
   const timeoutMs = transactionTimeoutMs(options.stage);
-  Object.assign(
-    record,
-    await waitForReceiptLifecycle(options.provider, transactionHash, Date.parse(record.submittedAt!), timeoutMs, rpc),
-  );
+  const [lifecycle] = await Promise.all([
+    waitForReceiptLifecycle(options.provider, transactionHash, Date.parse(record.submittedAt!), timeoutMs, rpc),
+    settleQuietly(submission.confirmed),
```
djizus: **P1 — Apply the transaction deadline to the Herald wait too.** `Promise.all` cannot finish until `submission.confirmed` settles, but `GameSyncRuntime.waitForTransaction` has no timeout. If Herald never delivers the status, the RPC receipt loop can return `confirmation_timeout` while this operation still waits indefinitely, preventing the workload from producing a report.

Reproduced with an unresolved `confirmed` promise and a provider returning `RECEIVED`: advancing the receipt clock past the 30-second workload deadline leaves `trackTransaction` pending. Bound the complete operation, report the missing stream confirmation explicitly, and stop further planning against stale state.

apps/agent-runner/src/tools/act.ts
```diff
@@ -0,0 +1,203 @@
...
+      ? armyPaths(client, params.explorerId as ID)
+      : structurePaths(client, params.structureId as ID);
+  return { kind: "planned", paths: [...paths.getPaths().values()].map(describePath) };
+};
+
+const armyPaths = (client: GameClient, explorerId: ID): ActionPaths => {
+  const { currentDefaultTick, currentArmiesTick } = getBlockTimestamp();
+  return client.actions.armyPaths({
+    explorerId,
+    ...buildArmyPathIndexes(client),
```
djizus: **P2 — Select path indexes using the explorer's layer.** Omitting `alt` makes `buildArmyPathIndexes` use its surface default (`false`). After spire travel, the army manager derives Ethereal movement from the explorer row but receives surface occupancy and explored tiles, so targets can be misclassified or rejected.

Reproduced by seeding an explorer with `coord.alt=true`, invoking `armyPaths` through `runAction`, and spying on the projection: it calls `getArmies(false)`. Read the explorer's authoritative layer and pass it through the related planning paths.

apps/agent-runner/scripts/play-blitz.sh
```diff
@@ -0,0 +1,104 @@
...
+      echo "play-blitz.sh: $name is required in the environment" >&2
+      exit 2
+    fi
+  done
+}
+
+# The newest open Blitz game on Herald, or the named one if it is open. Ended games are excluded by their clock, the
+# way the runner's phase check excludes them, so the run does not stop at its first tick.
+resolve_game_id() {
+  curl -fsS --max-time 15 "${VITE_PUBLIC_HERALD_URL%/}/madara/games" | python3 - "$GAME_NAME" <<'PY'
```
djizus: **P2 — Separate Python source from the piped JSON input.** The heredoc replaces the pipe as Python's stdin. `python3 -` consumes that stdin as the program, so `json.load(sys.stdin)` then sees EOF. The acceptance script exits with `JSONDecodeError` before starting a game even when curl returns valid Herald JSON.

Reproduced with curl stubbed to return a directory containing one valid live Blitz game; `resolve_game_id` exits 1 at `json.load`. Supply the program through `-c` or a script file, or pass the JSON through a separate descriptor.

djizus (changes requested):
The shared client builds and its existing tests pass, but four reproduced failures need fixing before merge: realm provisioning is skipped after settlement grants troops, the harness can wait indefinitely for Herald, Ethereal planning reads surface occupancy, and the full-Blitz gate script cannot read its JSON input. Details and reproduction conditions are attached inline.

Validation at this head: 274 core tests, 57 runner tests, 41 harness tests, runner typecheck, shared-package build, headless fixture smoke, and runner Docker build passed. The image's game, signer, and tool modules also imported with networking disabled. Two additional regression probes reproduced the provisioning and layer failures; separate probes reproduced the harness timeout and shell failures.

`pnpm run format` passed with no tracked changes. Local `pnpm run knip` could not complete because the filtered install omitted the app's `@eslint/js`; the remote knip check passed. Client CI is still failing its polling-discipline test on `packages/core/src/account/bind-gameplay-accounts.ts`: resolve the intended pre-session binding boundary and rerun the check.

The 96-bot acceptance run, full live Blitz agent game, and external sandbox/LLM gates remain unverified. The image build and offline tests do not clear those gates.

PR #4984, opened 2026-09-12
Add account notification preferences and confirmed local alerts

ponderingdemocritus:
Players can save an account notification level and opt this device into alerts for new confirmed activity involving them. Settings includes permission guidance, enable/disable controls, a test notification, and explicit loading/save/error states. Local alerts require a running game page; server push is follow-up work.

Herald preserves event confirmation updates. The client distinguishes confirmations after socket attachment from reconnect catch-up, uses event-time recipients, and groups mirrored battle/transfer records without collapsing distinct actions. Unknown story variants are excluded in production and diagnosed in development; policy inventory is checked against the generated StoryEvent manifest.

The service worker validates display envelopes and targets against the shared chain registry, deduplicates through IndexedDB, and suppresses local output if a browser push subscription exists. Clicks focus the matching game or open its entry flow without navigating an existing game's transactions.

The preference API is an Effect service with typed failures, session ownership checks and conditional revisions. `packages/db/src/schema/notifications.ts` is the only table definition. Apply it through the existing `pnpm --dir packages/db push` workflow against the identity database before deploying the API. There is no separate SQL migration.

The unused history cursor and contiguity migration are removed. Existing Herald progress markers remain trusted, so this change does not force a genesis backfill or make existing leaderboard/review data unavailable. The retired browser lifecycle runner and Playwright dependency are also removed, matching `next`.

Install Realms is now visible on landing pages and in game Settings. Chrome uses its saved native install prompt on click; Safari and other browsers show installation instructions. The control hides inside the installed app. Production assets and the manifest are verified by the build.

The production audit found that the Cloudflare custom domain changed `/sw.js` from the Pages origin's `no-cache` to `max-age=14400`. Deployment now maintains one scoped Cache Rule for the worker, manifest and offline page, preserving other rules, and verifies their public bytes, MIME types and revalidation headers after publication.

**Deployment prerequisite:** before merging, give the existing `CLOUDFLARE_API_TOKEN` Zone Read and Cache Rules Edit access for `realms.party`, retaining its Pages access. Missing permissions fail before publication. The rule automation is fixture-tested; the live zone has not been changed by this branch yet.

Successful account saves now invalidate preferences in other same-origin tabs. Those tabs pause local notification delivery while reloading the server's acknowledged preference. Invalidations arriving during a read/save trigger a fresh read after that operation settles, so a stale response cannot re-enable delivery. Failed refreshes remain visibly unsynced. Browser storage carries only the owner/revision invalidation, not the preference value.

Validation:
- Cross-tab review fix: 50 focused client tests, client typecheck, formatting and knip; two real Chromium tabs verified that an Off save makes the other tab transition through loading to Off without focus. Regression tests cover read/save races, account switches, publication errors and refresh failures.
- Latest install/cache changes: 40 focused client tests and four deployment/cache tests; production build and PWA budget verification (five entries, 38,918 bytes); browser fixture checks of mobile instructions and simulated native install events. Physical-device installation is still unverified.
- Earlier review fixes: 65 focused client tests; 34 focused core tests; nine policy/payload tests.
- 70 Herald tests, including a real PostgreSQL restart/checkpoint regression.
- 46 realms tests, including real PostgreSQL schema generation, preference races and typed failure checks.
- Client/Herald/realms typechecks, package builds, production client build, formatting, knip and whitespace checks.

Release limits: reconnect catch-up deliberately does not notify, even after a brief disconnect. Claims precede native display and can lose an alert after a crash/OS error; the design favors suppression of duplicates. Native API acceptance is not proof of an OS banner. Live-battle QA, physical Android/iOS checks and production-origin cookie verification remain open. Authenticated push subscriptions, VAPID and a durable notifier/outbox remain follow-up work; a future history cursor must land with its consumer.


djizus (changes requested):
I merged `next` into this branch (f46beb3) and fixed the one merge breakage (`live-world.test.ts` still expected the two-argument `appendEvents`). Format, knip, typecheck and the focused suites are green: client 190, core 97, Herald 98, notifications 8, realms 8. Postgres integration tests were not run locally.

What holds up: the cursor is monotonic and gap-safe, the preference API is session-authenticated with an owner precondition and revision conflicts, the claim-before-display IndexedDB dedup is atomic, and feed rows and audio cues now share one policy. Requesting changes for the items below.

**Blocking**
- The new Herald `GET /<chain>/history/story-events` cursor route (`apps/herald/src/http.ts`, `history-cursor.ts`, the `queryStoryHistory` path in `history-store.ts`) has no consumer anywhere in the repo; only its types and tests reference it. The notifier that would use it is described as follow-up work. Land it with its consumer or drop it from this PR; it is what drives the contiguity machinery and the rollout cost below.
- Rollout regression on the box: `HistoryStore.initialize()` adds `contiguous DEFAULT false`, so the existing progress marker is distrusted, `historyProgress()` returns null, and Herald re-backfills story history from genesis (500k+ blocks) over RPC. Until that completes the leaderboard stays not-ready and ended-game review throws "Herald history is complete through block none". Trust the marker written by the same single writer, or gate only the new endpoint on contiguity.
- The PWA lifecycle runner was removed from `deploy-client.yml` on `next` (it doubled deploy time). This branch still carries `verify-pwa-lifecycle.mjs`, `pwa-notification-checks.mjs`, the `playwright` dev dependency and the notifications.md paragraph about the runner. Drop them on the next merge with `next`.

**Should fix**
- `packages/notifications/src/story-policy.test.ts` and `delivery.test.ts` regex-read `events.cairo`, the battle emitter sources and `Scarb.lock`. That is the source-string test class we deleted on 2026-09-10; drive the policy inventory from the generated StoryEvent manifest instead.
- `apps/game/src/ui/features/event-feed/important-feed-rows.ts` now throws on a story name the policy table does not know (`story-policy.ts`). Before, an unknown variant was simply "not a battle". A Cairo story added before the table is updated crashes the feed and the audio-cue effect at render. Loud in dev, exclude in prod.
- Two sources of truth for the table: `apps/realms/server/migrations/001-notification-preferences.sql` and `packages/db/src/schema/notifications.ts`, while `packages/db` already applies schema with `drizzle-kit push`. Pick one for the box.
- `packages/notifications/src/delivery.ts` hardcodes `(madara|appchain)` in the target allowlist. `appchain` is dead and any new chain id fails validation, surfacing only as a settings error.
- `packages/core/src/sync/herald-game-sync-transport.ts`: `replayed` is true for every preconfirmed event and every confirmed event at or below the hello head, so nothing that happened during a disconnect is ever notified. Acceptable as an anti-flood choice; name it for what it is and document the gap in notifications.md.
- The new realms server code is plain async; the Effect standard for services is not advanced here.

**Rollout if merged**: apply the preferences table before the merge (deploy-box redeploys the identity server on push), stop the old Herald before the new one starts (single writer), and expect the backfill window above unless the marker is trusted.

PR #4994, opened 2026-09-14
Native Eternum: full game port and recorded sequencing randomness

djizus:
The native world supports persistent games, recorded execution and immutable presets. Gameplay now reads preset rules and a command mask; `mode_id` remains a label for consumers. The generated schema, bindings and client types follow the contract interfaces.

The pushed implementation includes Frontier FR3 through FR7:

- One season-long game contains daily epochs, separate exploration regions, persistent home economies and armies that expire at rollover.
- Presets control army limits, guards and attacks on owned sites. Captured camps pay the home realm, open a chest and refund the configured stamina. Captured rifts produce for the home realm until their cap or epoch end.
- Essence buys attunement. An army can enter an unlocked depth beside its daily home site for the depth's stamina cost.
- Reveals and captures resolve chests in the same action. Relics stay on the earning army. Cosmetic and token outcomes are recorded for later fulfillment, with a per-player, per-epoch token cap and relic pity by depth.

Combat, game records, relics, and realm configuration were moved along domain boundaries to keep every class below the unchanged 75,366-felt limit. The largest class at `d524c4d0d261` is RegistryDomain at 74,797 felts.

Validation covers epoch isolation and rollover, capture rewards and capped production, depth entry, chest distributions and claims, authorization, and the existing Blitz and Eternum rules. The native suite, schema drift check, deployer and protocol checks, client typechecks, formatting and knip pass locally. All hosted checks finished green on `d524c4d0d261`, including native-world, test-client, gallery, prettier and preferences-and-policy.

FR8 is uncommitted. Its board bonuses, actual-paid-labor demolition refunds and barracks upgrade flow pass the full native suite and JavaScript checks, but the wheat charge needs an owner ruling: removing production refills removes the existing wheat payment, while the design spec still prices troops at two wheat each. FR9, the fresh isolated deployment, FR11 and the Blitz free-slot regression remain outstanding. The Frontier client is outside this implementation pass.

The constant-time command-mask bit test reduced the same explore action from 174,280,023 to 174,273,906 L2 gas. This PR remains draft.


djizus:
Stream A has fast-forwarded `native-world-foundation` from `699102954e7` to `27eb16c9ffe2294e949de43d50c33da9253406c7`. The integration queue can resume.

[Exact-head validation](https://github.com/BibliothecaDAO/eternum/actions/runs/35958020663) is green. The final throwaway-shard smoke passed under the shared lock: game creation, recorded rejection with the exact domain reason, and Herald transaction/nonce delivery. Its containers and volumes are removed and the lock is released. No staging deployment.

djizus:
Stream E is preparing the accepted node-memory follow-up for landing from `27eb16c9ffe`. I will recheck `native-world-foundation` before the fast-forward and rebase/revalidate if F lands first. F: if E lands first, please rebase the two client commits onto that head. No deployments are part of this landing.

djizus:
Stream E update: `1d1f49ad62c` rebased cleanly onto `27eb16c9ffe`; every hosted job except native-world passed. Run 35962387562 reached native-world's 35-minute timeout while tests were still passing, with no test failure in its log. I am retrying only the incomplete job and the dependent validation check on the same head. `native-world-foundation` is still `27eb16c9ffe`; the second stream to land will rebase and revalidate.

djizus:
The E timeout was preceded by an approximately 18-minute cold build of corrected snforge. That binary was cached before the gates. E's retry has restored it and reached native-world tests on the same `1d1f49ad62c` head. F's first run now shows cancelled too; the same cached-binary retry may apply. Integration remains `27eb16c9ffe`.

djizus:
Stream E landed first: `native-world-foundation` fast-forwarded from `27eb16c9ffe` to `1d1f49ad62cb29713e68e6b09c9fa182b1a1f6c6` (accepted node-memory commit, unchanged patch after rebase).

[Exact-head validation 35962387562, attempt 2](https://github.com/BibliothecaDAO/eternum/actions/runs/35962387562/attempts/2) is green. The first attempt timed out after the cold snforge build; retry restored the cached binary. Local 15 Python checks, format and knip passed. No deployments.

F: rebase the two client commits onto `1d1f49ad62c` before the second landing, and validate that new head.

